PR #109, opened 2013-02-25
 tent/parallel

nno:
- added a module misc/parallelization as a wrapper around whatever kind of parallelization implementation might be available. 
- currently supports pprocess and has a fallback single process implementation that should always work.
- refactored Searchlight and surface-based voxel selection code
- added parallelization support to RepeatedMeasure


bpinsard:
I would be greatly interested by parallelization of RepeatedMeasure. What is the status of this PR? Thanks

PR #235, opened 2014-07-15
Make SplitRFE work with "static" univariate measures + BinaryFxFeaturewiseMeasure

yarikoptic:
The purpose was to unify/simplify feature selection based on some univariate measures with nested cross-validation  to estimate optimal number of features. Previously SplitRFE was allowing only for using classifier's sensitivity analyzer, and not some external (possibly not changing through RFE, thus "static") measure.  With these changes it should be easy to wrap any binary function to estimate relation to the targets samples attribute (e.g. mutual information, or correlation), and provide it to SplitRFE.  
SplitRFE also got ability to parallelize its inner loop via joblib (well -- I think eventually we might just shift to use it for all parallelization altogether, but not yet).
I forgot if there was anything particular not finished (besides those various sandboxy TODOs), and it seemed to function nicely, so here came a PR ;)


yarikoptic:
@hanke  should we may be get this one into the release as well?

hanke:
@yarikoptic Looks good, but it would be awesome to have a tiny example that documents this enhancement.

yarikoptic:
indeed, a little example of SplitRFE function could be utterly useful, but not sure if I would pull it off... where do you think it should/would go besides the docstring?  we have   doc/source/classifiers.txt and doc/source/featsel.txt but are we using it?  Ideally there should be an example demonstrating this beast.

yarikoptic:
ok -- added some example(s) ;)

yarikoptic:
grrr... where is this failure coming from??? can't replicate locally in precise environemnt... I guess would need to backdoor again to figure it out :-/

yarikoptic:
Ha -- I've fixed it! @PyMVPA/developers thus ready for the review -- tests now pass ;)

PR #259, opened 2014-12-11
OpenFMRI enhancement

hanke:
Make per-trial modeling simpler.


hanke:
Just to get the ball rolling...

hanke:
@nno @yarikoptic I intent to merge this one too. I removed all the bits that were still in a "quick-hack" stage, but the remaining core bits are used quite heavily over here already.

This class will need to evolve with OpenFMRI, but there is enough demand to make it worthwhile.

Please veto if you must.

nno:
@hanke looks good to me.

hanke:
@nno Thanks! Merging.

PR #267, opened 2015-01-11
ENH: at least if we still convert to python3 without problems

yarikoptic:
Let's wait for travis!


yarikoptic:
travis was happy but I was forgetful .  Merging now!

PR #269, opened 2015-01-21
ENH: "chunk" strategy for AttributePermutator

hanke:


mvpa2/generators/permutation.py
```diff
@@ -58,14 +58,16 @@ def __init__(self, attr, count=1, limit=None, assure=False,
...
           'simple' strategy is the straightfoward permutation of attributes (given
           the limit).  In some sense it assumes independence of those samples.
           'uattrs' strategy looks at unique values of attr (or their unique
           combinations in case of `attr` being a list), and "permutes" those
           unique combinations values thus breaking their assignment to the samples
           but preserving any dependencies between samples within the same unique
-          combination.
+          combination. The 'chunks' strategy swaps attribute values of entire chunks.
+          Naturally, this will only work if there is the same number of samples in
+          all chunks.
```
yarikoptic: and if samples were not  reordered/sorted  like eg fxmapper  used to do it

mvpa2/generators/permutation.py
```diff
@@ -39,7 +39,7 @@ class AttributePermutator(Node):
     dataset.
     """
     def __init__(self, attr, count=1, limit=None, assure=False,
-                 strategy='simple', rng=np.random, **kwargs):
+                 strategy='simple', chunk_attr=None, rng=np.random, **kwargs):
```
yarikoptic: originally we used limit  to define grouping  here,  could/ should  it be used here as well  or you see a  use case  where  both would need to be specified

mvpa2/generators/permutation.py
```diff
@@ -94,6 +97,10 @@ def _get_pcfg(self, ds):
         return get_limit_filter(self._limit, collection)
 
 
+    def get_chunk_attr(self, ds):
```
yarikoptic: get  usually  returns  something...  better be protected  anyways

mvpa2/generators/permutation.py
dinga92: I am not exactly sure what you mean, do you mean case for example after hrf modeling, you'll end up with one sample per condition and without order within chunks, so every chink have samples in the same order? In that case swapping labels by that strategy wouldn't do anything

mvpa2/generators/permutation.py
dinga92: I don't see such a case. But limit ant chunk_attr does different things. If you specify limit, then it will permute labels only within limit and not between (am I right here?) for chunk strategy it is opposite, you will not permute within chunk_attr, but you will swap around chunk labels in original order between chunks. Also it's doing different thing in the code -> that get_chunk_attr function

mvpa2/generators/permutation.py
dinga92: like what? that function is here because  _permute* functions doesn't have access to ds, so you cannot get chunk attributes directly or call that function within _permute* to return chunk attributes. It deserve a comment anyway

mvpa2/generators/permutation.py
yarikoptic: yes -- that is the kind of cases I had in mind.
For that reason it might also be nice to add a sanity check, that not all labels across all chunks are the same.

mvpa2/generators/permutation.py
yarikoptic: like ```_assign_chunk_attr```. I would have different approach though to make it more explicit... easier to do than to say -- I will send you a PR shortly -- you decide either to accept ;)

yarikoptic: not sure what is the workflow between you and @hanke but now he got a PR https://github.com/hanke/PyMVPA/pull/8 ;-)

hanke:
@yarikoptic this looks ready to merge. Veto?

yarikoptic:
Sure -- could be merged, but better still gets the check/warning if all orders are the same across chunks.

yarikoptic:
ok -- let me merge and add a simple warning fist

PR #270, opened 2015-01-22
NF: Added an option to include center feature in HollowSphere.

swaroopgj:
Added 'include_center' option HollowSphere to include center feature's data in the dataset sent to measures as first feature.

### DIFF
--- a/mvpa2/misc/neighborhood.py
+++ b/mvpa2/misc/neighborhood.py
@@ -232,7 +232,7 @@ class HollowSphere(Sphere):
 
     """
 
-    def __init__(self, radius, inner_radius, **kwargs):
+    def __init__(self, radius, inner_radius, include_center=False, **kwargs):
         """ Initialize the Sphere
 
         Parameters
@@ -245,6 +245,9 @@ class HollowSphere(Sphere):
           Inner radius of the 'sphere', describing where hollow
           part starts.  It is inclusive, so `inner_radius` of 0,
           would already remove the center element.
+        include_center : bool
+          Flag indicating whether to include the center element.
+          Center element is added as first feature. (Default: False)
         **kwargs
           See `Sphere` for additional keyword arguments
         """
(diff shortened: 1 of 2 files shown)


mvpa2/misc/neighborhood.py
```diff
@@ -296,7 +299,7 @@ def _get_increments(self, ndim):
 
         if not len(res):
             warning("%s defines no neighbors" % self)
-        return res
+        return np.vstack([np.zeros(ndim,dtype='int'),res]) if self.include_center else res
```
yarikoptic: you never assign it to self.  So it leads to tests failures:  https://travis-ci.org/PyMVPA/PyMVPA/jobs/47877969  and for proper this PR should have a test assigned (just set this parameter and verify that you do get center reported, and that it is first as you promised)

swaroopgj:
Fixed it and added unit test.
Test passes locally.

mvpa2/misc/neighborhood.py
```diff
@@ -253,6 +256,7 @@ def __init__(self, radius, inner_radius, **kwargs):
                   "than the radius (got %g)" % (inner_radius, radius)
         Sphere.__init__(self, radius, **kwargs)
         self._inner_radius = inner_radius
+        self.include_center = include_center
```
yarikoptic: now also adjust ```__repr__``` below to add ```include_center ``` when it is not default... check other repr's for examples

swaroopgj:
I added it to include_center to __repr__
Hopefully the right way :-|

mvpa2/misc/neighborhood.py
```diff
@@ -253,10 +256,11 @@ def __init__(self, radius, inner_radius, **kwargs):
...
 
     def __repr__(self, prefixes=[]):
         return super(HollowSphere, self).__repr__(
-            ['inner_radius=%r' % (self._inner_radius,)])
+            ['inner_radius=%r' % (self._inner_radius,),'include_center=%r' % (self.include_center,)])
```
yarikoptic: you did it almost right ;-)  so I will merge it and tune it up ... if you use

```Python
+ _repr_attrs(self, ['include_center'], default=False))
```
then it wouldn't even list if it it matches the default

PR #271, opened 2015-01-27
noncontiguous  splits samples generator

bpinsard:
see the doc for further description
do not know if this is the optimal or recommended coding, waiting for ideas/comments.

### DIFF
--- a/mvpa2/generators/resampling.py
+++ b/mvpa2/generators/resampling.py
@@ -203,3 +203,61 @@ class Balancer(Node):
     def __str__(self):
         return _str(self, str(self._amount), n=self._attr, count=self.count,
                     apply_selection=self._apply_selection)
+
+
+class NonContiguous(Node):
+    """Generator to remove samples too close as measured with a sample attribute
+    eg.:
+    >>> partitioner = ChainNode([NFoldPartitioner(),
+    NonContiguous(attr='time',
+    dist=60,
+    partition_keep=2,
+    partition_trim=1)])
+    if sa.time in sec then all the training samples in partition(_keep)=1
+    that are less than 60 secs distant from partition(_trim)=2 
+    will be assigned to partition 0 to be excluded from cross-validation fold
+    with appropriate splitter
+    """
+    def __init__(self,
+                 attr='chunks',
+                 dist=1,
+                 partitions_attr='partitions',
+                  partition_keep=2,
+                 partition_trim=1,
+                 **kwargs):
+        """
+        Parameters
+        ----------
+        dist : the minimum distance between the samples to the 2 splits of data
+        partitions_attr : the attribute describing the split
+        partition_keep : the partition that is to be kept
+        partition_trim : the partition to be trimmed
+        """
+        Node.__init__(self, **kwargs)
+        self._attr = attr
+        self._dist = dist
+        self.partitions_attr = partitions_attr
+        self.partition_keep = partition_keep
+        self.partition_trim = partition_trim
+        
+    def _call(self,ds):
+        attr, collection = ds.get_attr(self._attr)
+        orig_partitioning = ds.sa[self.partitions_attr].value
+        
+        attr_keep = np.unique(attr[orig_partitioning==self.partition_keep])
+        trim_mask = orig_partitioning==self.partition_trim
+        new_partitioning = orig_partitioning.copy()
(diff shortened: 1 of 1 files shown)


mvpa2/generators/resampling.py
```diff
@@ -203,3 +203,71 @@ def generate(self, ds):
...
+
+
+class NonContiguous(Node):
+    """Generator to remove samples too close as measured with a sample attribute
+    eg.:
+    >>> partitioner = ChainNode([NFoldPartitioner(),
+    NonContiguous(attr='time',
+    dist=60,
+    partition_keep=2,
+    partition_trim=1)])
```
yarikoptic: please reformat using ReST formatting, otherwise would not look good on the web.  Checkout other docstrings for examples

mvpa2/generators/resampling.py
```diff
@@ -203,3 +203,71 @@ def generate(self, ds):
...
+                 **kwargs):
+        """
+        Parameters
+        ----------
+        dist_attr : the attribute to measures distance
+        dist : the minimum distance between the samples to the 2 splits of data
+        split_attr : the attribute describing the split
+        split_keep : the list of splits that are to be kept : default [2] (testing)
+        split_trim : the list of splits to be trimmed : default [1] (training)
+        split_assign : the value to be assigned to trim samples
```
yarikoptic: same here -- ideally should follow the NumPy convention for parameters description here.  Just see other docstrings in the code

yarikoptic:
code without unittests is broken code ;)

mvpa2/generators/resampling.py
```diff
@@ -203,3 +203,74 @@ def generate(self, ds):
     def __str__(self):
         return _str(self, str(self._amount), n=self._attr, count=self.count,
                     apply_selection=self._apply_selection)
+
+
+class NonContiguous(Node):
+    """Generator to remove samples too close as measured with a sample attribute
+    ``` partitioner = ChainNode([NFoldPartitioner(),
```
yarikoptic: We use numpy style docstrings, and run doctests which allows to assure that code we inject is correct.  Ideally such a snippet should be entered in the same fashion, see various places around the code (where there is Examples in docstrings) .e.g  within mvpa2/base/dataset.py

mvpa2/generators/resampling.py
```diff
@@ -203,3 +203,74 @@ def generate(self, ds):
...
+                 split_assign=[3],
+                 **kwargs):
+        """
+        Parameters
+        ----------
+        dist_attr : str
+          the attribute to measures distance
+        dist : all type of value that can be compared to dist_attr
+          the minimum distance between the samples to the 2 splits of data
+        split_attr : str
```
yarikoptic: boring but ```, optional``` ;-)

mvpa2/tests/test_generators.py
```diff
@@ -248,6 +248,18 @@ def test_balancer():
...
+    dist = 2
+    chunk = 4
+    nc = NonContiguous(dist=2)
+    ds.sa['partitions'] = (ds.chunks==chunk)+1
+    res = nc(ds)
+    trim_mask = np.logical_and(ds.sa.partitions==1, np.abs(ds.chunks-chunk)<=dist)
+    assert_array_equal( res.sa.partitions[trim_mask], 3)
+    assert_array_equal( res.sa.partitions[np.logical_not(trim_mask)],
+                        ds.sa.partitions[np.logical_not(trim_mask)])
+
```
yarikoptic: given the amount of options I feel that one testing scenario would be too optimistic ... could you come up with at least one which uses different from default options for which partitions to check etc? 

while at it I would appreciate if you could prettify the formatting a bit so it gets closer to recommended by PEP8. (e.g. no spaces after ```(``` in above.  I would recommend PyCharm IDE -- really neat one.  In NeuroDebian we have now pycharm-community-sloppy   package so you could just apt-get install it happen you use NeuroDebian.

Cheers and thanks for all your work!!!

yarikoptic:
@bpinsard interested to work on this one a bit further? (comments were posted a bit ago)

hanke:
So what is the state of things? Having the doc snippet as a test would be nice, but OTOH it doesn't do something real anyway. Can we merge this before the 12 months mark? ;-)

PR #272, opened 2015-01-27
fix for splitting with excluded partition + subsampling dataset generator

bpinsard:
trying to fix problems I had with using a generator that subsamples the dataset (but does not changes the targets) plus had some excluded partitions,
maybe the splitter could be a parameter of the searchlight?


mvpa2/measures/adhocsearchlightbase.py
```diff
@@ -346,11 +346,11 @@ def _sl_call(self, dataset, roi_ids, nproc):
         # indicies
         # XXX we could make it even more lightweight I guess...
         dataset_indicies = Dataset(np.arange(nsamples), sa=dataset.sa)
-        splitter = Splitter(attr=generator.get_space())
+        splitter = Splitter(attr=generator.get_space(),attr_values=[1,2])
```
yarikoptic: minor: space after ```,``` -- pep8

yarikoptic:
interesting! thanks. 

But for exclude you better should have used e.g. partition "3" (or 10 for what I care ;) ) thus leaving default 1, 2 for train/test as it is now.  We already do assign partition to 3 in some partitioners I believe when we would like to exclude those samples from training/testing splits.  That would eliminate the need for tuning splitter ere.

Also ideally we need a unittest which would demonstrate presence of problem before and absence of it after fix ;) (for the 2nd change of taking indicies for labels).  I guess if you created some simple CustomPartitioner without waiting for the other PR to get accepted -- would be the best

bpinsard:
ok thanks for your fast comments!
I didnt't knew that the default splitting would be 0,1,2 only, will change for 3 the default assigned value in the other PR and reverse the Splitter change.
will try to write relevant test

mvpa2/measures/adhocsearchlightbase.py
```diff
@@ -346,11 +350,14 @@ def _sl_call(self, dataset, roi_ids, nproc):
         # indicies
         # XXX we could make it even more lightweight I guess...
         dataset_indicies = Dataset(np.arange(nsamples), sa=dataset.sa)
-        splitter = Splitter(attr=generator.get_space())
+        if self._splitter is None:
+            splitter = Splitter(attr=generator.get_space())
+        else:
+            splitter = self._splitter
```
yarikoptic: I believe could be a bit more concise with
```splitter = self._splitter or Splitter(attr=generator.get_space())```

File: mvpa2/measures/adhocsearchlightbase.py [hunk 3]
```diff
@@ -136,6 +136,7 @@ class SimpleStatBaseSearchlight(BaseSearchlight):
     def __init__(self, generator, queryengine, errorfx=mean_mismatch_error,
                  indexsum=None,
                  reuse_neighbors=False,
+                 splitter=None,
```
yarikoptic: please add a corresponding entry to ```__repr__```

yarikoptic:
```> will try to write relevant test```
I don't see any test added to test giving another splitter here

hanke:
@yarikoptic @bpinsard What are your plans on this one. Can you fix it up for a quick merge. The release is imminent and it would be nice to see this resolved. It seems pretty much ready.

hanke:
@yarikoptic @bpinsard Little ping. Last comments were minor. Merge?

yarikoptic:
I hoped to get a   tiny unittest  for this  additional  argument

PR #274, opened 2015-02-11
ENH Add identity neighborhood

effigies:
Neighborhood such that `neighbors(i) == [i]`.

Perhaps there already exists a way of doing this (apart from, say, `Sphere(0.5)`), but since it was easier for me to write this than dig deeper into the code, I figured it might be useful for others to have something big and obvious.

### DIFF
--- a/mvpa2/misc/neighborhood.py
+++ b/mvpa2/misc/neighborhood.py
@@ -30,6 +30,7 @@ class IdentityNeighborhood(object):
     """
 
     def __init__(self):
+        """Initialize the neighborhood"""
         pass
 
     def __repr__(self):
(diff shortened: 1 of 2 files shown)


nno:
Thanks for the idea, but two issues:
1) Can you add a unit test? Probably best is in mvpa2/tests/test_neighborhood.py
2) Travis complaints:
FAIL: test_docstrings (mvpa2.tests.test_suite.SuiteTest)
  File "/home/travis/build/PyMVPA/PyMVPA/mvpa2/tests/test_suite.py", line 96, in test_docstrings
    self.fail('\n'.join(sfailures))
AssertionError: Some items have missing docstrings:
 types: IdentityNeighborhood.__init__

yarikoptic:
hint: the simplest unittest would be to verify that output is identical to that ```Sphere(0.5)``` ;)

PR #277, opened 2015-02-19
BF: do not demean in procrustean if demean=False (Close #276)

### LINKED ISSUE #276: ProcrusteanMapper(demean=False) what should happen
@swaroopgj I am seeing this

```
  File "..../mvpa2/mappers/procrustean.py", line 96, in _train
    data = data - mean
TypeError: unsupported operand type(s) for -: 'float' and 'NoneType'
```

when I train `ProcrusteanMapper(demean=False)`. Looking into ways of fixing this, I am uncertain what should happen when this argument is given.

The way the code is written, not demeaning is not an option. Should it be changed to allow for this, or should setting demean=False raise an error?

Thanks in advance.

yarikoptic:
although probably doesn't make algorithmic sense it is probably the cleanest way to conform to super class'es desire to have demean option

### DIFF
--- a/mvpa2/mappers/procrustean.py
+++ b/mvpa2/mappers/procrustean.py
@@ -89,11 +89,15 @@ class ProcrusteanMapper(ProjectionMapper):
                 data = ds
             if assess_residuals:
                 odatas += (data,)
-            if i == 0:
-                mean = self._offset_in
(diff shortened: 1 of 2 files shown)


nno:
Thanks, this looks good to me.

PR #279, opened 2015-02-23
fix for one-class (GLQ)DA

bpinsard:
squeezing too far when dataset is one-class.

### DIFF
--- a/mvpa2/clfs/dummies.py
+++ b/mvpa2/clfs/dummies.py
@@ -28,7 +28,7 @@ class SameSignClassifier(Classifier):
     """Dummy classifier which reports +1 class if both features have
     the same sign, -1 otherwise"""
 
-    __tags__ = ['notrain2predict', 'one-class-binary']
+    __tags__ = ['notrain2predict', 'oneclass-binary']
     def __init__(self, **kwargs):
         Classifier.__init__(self, **kwargs)
 
@@ -52,7 +52,7 @@ class RandomClassifier(Classifier):
     """Dummy classifier deciding on labels absolutely randomly
     """
 
-    __tags__ = ['random', 'non-deterministic', 'one-class-binary']
+    __tags__ = ['random', 'non-deterministic', 'oneclass-binary']
 
     same = Parameter(
         False, constraints='bool',
--- a/mvpa2/clfs/dummies.py
+++ b/mvpa2/clfs/dummies.py
@@ -28,7 +28,7 @@ class SameSignClassifier(Classifier):
     """Dummy classifier which reports +1 class if both features have
     the same sign, -1 otherwise"""
 
-    __tags__ = ['notrain2predict']
+    __tags__ = ['notrain2predict', 'one-class-binary']
     def __init__(self, **kwargs):
         Classifier.__init__(self, **kwargs)
 
@@ -52,7 +52,7 @@ class RandomClassifier(Classifier):
     """Dummy classifier deciding on labels absolutely randomly
     """
 
-    __tags__ = ['random', 'non-deterministic']
+    __tags__ = ['random', 'non-deterministic', 'one-class-binary']
 
     same = Parameter(
         False, constraints='bool',
--- a/mvpa2/clfs/gda.py
+++ b/mvpa2/clfs/gda.py
@@ -95,7 +95,7 @@ class GDA(Classifier):
             raise ValueError, \
                   "No idea on how to handle '%s' way to compute priors" \
                   % self.params.prior
-        return priors
+        return np.atleast_1d(priors)
 
 
     def _train(self, dataset):
--- a/mvpa2/clfs/gda.py
+++ b/mvpa2/clfs/gda.py
@@ -47,7 +47,7 @@ class GDA(Classifier):
 
     """
 
-    __tags__ = ['binary', 'multiclass', 'one-class']
+    __tags__ = ['binary', 'multiclass', 'oneclass']
 
 
     prior = Parameter('laplacian_smoothing',
--- a/mvpa2/clfs/gda.py
+++ b/mvpa2/clfs/gda.py
@@ -47,7 +47,7 @@ class GDA(Classifier):
 
     """
 
-    __tags__ = ['binary', 'multiclass']
+    __tags__ = ['binary', 'multiclass', 'one-class']
 
 
     prior = Parameter('laplacian_smoothing',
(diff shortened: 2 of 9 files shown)


yarikoptic:
could you add a unittest which would fail with previous state of affairs but now succeed?

bpinsard:
sure. is there a file with LDA tests or is it factored for all classifiers?

yarikoptic:
On Mon, 23 Feb 2015, bpinsard wrote:

>    sure. is there a file with LDA tests or is it factored for all
>    classifiers?

I guess it is  factored... feel welcome to add to test_clf as a new test
and import LDA directly if necessary and use sweepargs if you need
to test multiple)

bpinsard:
it seems the relevant test was just skipped for GDA, but sklearn GDA doesn't pass it.
added skl to sweepargs.

yarikoptic:
hm, but that would ignore all sklearn classifiers, even those which passed before... a bit confused

bpinsard:
true, can we use combination of tags for sweepargs?
or is there a tags more specific to exclude just skl lda?
or should we tag the classifiers able to do one-class learning/prediction with a new tag?

yarikoptic:
there definetely should be a tag for one-class...  I could swear we already have it but looking into warehouse.py -- apparently not.  Please feel more than welcome to add ;-) !

bpinsard:
there is this tag here:
https://github.com/PyMVPA/PyMVPA/blob/ae9efefdf0265b6f48821031f58766181f68b86c/mvpa2/clfs/libsvmc/svm.py#L67
but never used in warehouse.
However what falls under this tag?
Some classifier will accept one-class dataset and provide +1 -1 results other will always predict the only label they learned, but some of them will produce estimates/probabilities that can be of interest.

yarikoptic:
>    there is this tag here:
>    https://github.com/PyMVPA/PyMVPA/blob/ae9efefdf0265b6f48821031f58766181f68b86c/mvpa2/clfs/libsvmc/svm.py#L67
>    but never used in warehouse.

indeed, and that is why not listed in the warehouse...

>    However what falls under this tag?
>    Some classifier will accept one-class dataset and provide +1 -1 results
>    other will always predict the only label they learned, but some of them
>    will produce estimates/probabilities that can be of interest.

good question ;)  May be then two tags? 

"oneclass-binary" for the former (+1 -1) and "oneclass" for the later (always "+1")?

We never really tagged (yet) for either  estimates/probabilities
are produced and in what form... could be the next step ;)

yarikoptic:
On Mon, 23 Feb 2015, bpinsard wrote:

>    true, can we use combination of tags for sweepargs?
>    or is there a tags more specific to exclude just skl lda?
I don't think we provisioned exclusion of combinations of tags.  Could
also be done though with a small "fix" if you like ;)

mvpa2/clfs/gda.py
```diff
@@ -47,7 +47,7 @@ class GDA(Classifier):
 
     """
 
-    __tags__ = ['binary', 'multiclass']
+    __tags__ = ['binary', 'multiclass', 'one-class']
```
yarikoptic: since we have already 'multiclass', then may be 'oneclass'? ;)

mvpa2/clfs/dummies.py
```diff
@@ -52,7 +52,7 @@ class RandomClassifier(Classifier):
     """Dummy classifier deciding on labels absolutely randomly
     """
 
-    __tags__ = ['random', 'non-deterministic']
+    __tags__ = ['random', 'non-deterministic', 'one-class-binary']
```
yarikoptic: I wonder though why I didn't give it even "binary" or "mutliclass" originally... hm...  probably so it doesn't even participate in any case where meaningful classifier is requested (e.g. in tests), so probably not worth adding oneclass tags here either

mvpa2/tests/test_clf.py
```diff
@@ -335,7 +335,8 @@ def test_degenerate_usage(self, clf):
 
 
     # TODO: sg - remove our limitations, meta, lda, qda and skl -- also
-    @sweepargs(clf=clfswh['!sg', '!plr', '!meta', '!lda', '!qda', '!glmnet'])
+    @sweepargs(clf=clfswh['one-class', '!sg', '!plr', '!meta', '!glmnet'] + \
+                   clfswh['one-class-binary', '!sg', '!plr', '!meta', '!glmnet'])
```
yarikoptic: I guess with introduction of those tags it could simply be `['oneclass', 'oneclass-binary']` without any ugly exclusions? (may be  with '!meta' at most)?

mvpa2/tests/test_clf.py
```diff
@@ -343,6 +344,7 @@ def test_single_class(self, clf):
         ds = ds[ds.sa.targets == 'L0']  #  only 1 label
         assert(ds.sa['targets'].unique == ['L0'])
 
+        print clf
```
yarikoptic: at the end there should be no prints in the tests

you can use environ variable  MVPA_DEBUG=TEST   to enable printout of args of running sweep_args decorator ;)

mvpa2/clfs/dummies.py
bpinsard: it' not clear for me if the tag means that the classifier is supposed not to crash when given a one-class dataset, or if it makes any sense to be trained with such?

yarikoptic: >    it' not clear for me if the tag means that the classifier is supposed not
>    to crash when given a one-class dataset, or if it makes any sense to be
>    trained with such?

more of a sense than tolerance indeed ;)

mvpa2/tests/test_clf.py
bpinsard: ```$MVPA_DEBUG=TEST nosetests mvpa2/tests/test_clf.py```
doesn't display the args =[

mvpa2/clfs/warehouse.py
```diff
@@ -47,7 +47,7 @@
         'libsvm', 'sg', 'meta', 'retrainable', 'gpr',
         'notrain2predict', 'ridge', 'blr', 'gnpp', 'enet', 'glmnet',
         'gnb', 'plr', 'rpy2', 'swig', 'skl', 'lda', 'qda',
-        'random-forest', 'extra-trees', 'random']
+        'random-forest', 'extra-trees', 'random', 'oneclass', 'oneclass-binary']
```
yarikoptic: sorry for a delay with the review and thanks for the contribution!! would you mind just adding 1 little clarification here in the comment by each "oneclass" and "...-binary" of what is the difference there ;)

hanke:
@bpinsard Please let me know your thoughts on @yarikoptic last request. I can add the comment myself, in the interest of a speedy merge. Thanks.

PR #281, opened 2015-03-04
fix label bug in TreeClassifier

bpinsard:
truncated labels occurs in case of labels being string array with variable length and first predicted is shorter
guessing dtype from learned labels set instead

### DIFF
--- a/mvpa2/tests/test_clf.py
+++ b/mvpa2/tests/test_clf.py
@@ -1093,7 +1093,40 @@ class ClassifiersTests(unittest.TestCase):
             xx = pl.hist(accs, bins=bins, align='left')
             pl.xlim((0. - step/2, 1.+step/2))
 
+    @sweepargs(clf=clfswh['multiclass'])
+    def test_diff_len_labels_str(self, clf):
+        # check if the classifier can handle a dataset with labels as string of
+        # variable length
+        # was failing on TreeClassifier due to np.str dtype being assumed from first
+        # returned value
+        ds = datasets['uni4small'].copy()
+        newlabels = dict([(l,l+'_'*li) for li,l in enumerate(ds.uniquetargets)])
+        ds.targets = [newlabels[l] for l in ds.targets]
 
+        clf2 = clf.clone()
+        clf2.train(ds)
+        predictions = clf2.predict(ds)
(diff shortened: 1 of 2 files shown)


nno:
- can you provide a unit test, preferably one that fails with the old code but not with this change?
- trained_targets is a conditional attribute. Is it possible that it is not present in the dataset (it seems so)? In that case the code would fail.

bpinsard:
- trained_targets is a ca of the classifier, not the dataset, and should be generated at training (unless specifically deactivated), am I right?
- I am suprised that the test doesn't already include a case with string labels, but it might be skipped for meta classifiers. Can try to write one, but should it be more general than just TreeClassifier?

nno:
- You're right about the trained_targets being a .ca of the classifier, sorry about that. In base/learner.py (which provides Learner, which is a superclass of TreeClassifier) it is set to True as default. I just wonder what would happen if it is disabled?
- Writing a test for just TreeClassifier would be a good start, but if you would also be willing to write  a more general one, that would be great!

mvpa2/tests/test_clf.py
```diff
@@ -1093,7 +1093,40 @@ def test_gideon_weird_case(self):
...
+        # returned value
+        ds = datasets['uni4small'].copy()
+        newlabels = dict([(l,l+'_'*li) for li,l in enumerate(ds.uniquetargets)])
+        ds.targets = [newlabels[l] for l in ds.targets]
 
+        clf2 = clf.clone()
+        clf2.train(ds)
+        predictions = clf2.predict(ds)
+        # predictions on the same ds as training should give same labels
+        assert(np.all(np.unique(predictions) == ds.uniquetargets))
```
yarikoptic: generally it is not really has to be true.  it is ok to have training error, otherwise we would be overfitting.  if there is training error, we might miss one label completely among predictions -- it is unlikely but possible.  or am I missing smth?

yarikoptic: so what about testing smth like   ```set(ds.uniquetargets).issuperset(predictions)```?

yarikoptic: just a note -- if we were to do this np.all comparison, use ```assert_array_equal```

PR #284, opened 2015-04-15
Group accuracy cluster evaluation after Stelzer et al (improved)

hanke:
This is not yet intended to be merged, but it is complete enough to give a good impression of the API and the functionality. Please comment.

@dinga92 I started the main PR to eventually get this into the main line 

TODO:
- Few missing tests detailed at the end of the unit test code
- needs an example in the docs. however I am undecided whether to prep a new dataset from real data, or generate some artificial numbers....


nno:
Nice work.

Any ideas about how easy or difficult it would be to add support for:

(a) threshold-free cluster enhancement (TFCE) [1] - this avoids the researcher's degree of freedom for setting an uncorrected threshold, is a more continuous measure, and have a few more advantages.
(b) datasets with arbitrary neighbourhoods [2], or at least surface-based datasets (it seems that currently only volumetric array-like datasets are supported).

In CoSMoMVPA there is support for both of these features [3], but I don't know how easy it would be to port that code... I had a quick glance at the current code; re (a), one value is stored for each feature rather than each cluster. re (b), this may potentially be achieved through subclassing QueryEngine, e.g. a subclass ClusteringQueryEngine with, for now, sub-subclasses VolumetricClusteringQueryEngine and SurfaceClusteringQueryEngine.

[1] http://www.fmrib.ox.ac.uk/analysis/techrep/tr08ss1/tr08ss1.pdf
[2] https://github.com/CoSMoMVPA/CoSMoMVPA/blob/master/mvpa/cosmo_cluster_neighborhood.m
[3] https://github.com/CoSMoMVPA/CoSMoMVPA/blob/master/mvpa/cosmo_montecarlo_cluster_stat.m

mvpa2/algorithms/group_clusterthr.py
```diff
@@ -0,0 +1,427 @@
...
+    from mvpa2.base import debug
+
+import os
+import random
+import bisect
+from collections import Counter
+
+import numpy as np
+
+from scipy.ndimage import measurements
```
yarikoptic: shouldn't we also get import in the mvpa2.suite  guarded for availability of scipy?

mvpa2/algorithms/group_clusterthr.py
```diff
@@ -0,0 +1,427 @@
...
+    group-average accuracy maps under the NULL hypothesis, as described in [1]_.
+
+    Note: this class implements a modified version of that algorithm. The
+    present implementation differs in, at least, three aspects from the
+    description in that paper.
+
+    1) Cluster p-values refer to the probability of observing a particular
+       cluster size or a larger one (original paper: probability to observe a
+       larger cluster only).  Consequently, probabilities reported by this
+       implementation will have a tendency to be higher in comparison.
```
yarikoptic: wouldn't it be easy to provide option to perform as in original publication as well?  would ease comparison etc?

mvpa2/algorithms/group_clusterthr.py
```diff
@@ -0,0 +1,427 @@
...
+    Instances of this class must be trained before than can be used to
+    threshold accuracy maps. The training dataset must match the following
+    criteria:
+
+    1) For every subject in the group, it must contain multiple accuracy maps
+       that are the result of a within-subject classification analysis
+       based on permuted class labels. One map must corresponds to one fixed
+       permutation for all features in the map, as described in [1]_. The
+       original authors recommend 100 accuracy maps per subject for a typical
+       searchlight analysis.
```
yarikoptic: I have this vague memory of 20 being already reasonable...?

mvpa2/algorithms/group_clusterthr.py
```diff
@@ -0,0 +1,427 @@
...
+    References
+    ----------
+    .. [1] Johannes Stelzer, Yi Chen and Robert Turner (2013). Statistical
+       inference and multiple testing correction in classification-based
+       multi-voxel pattern analysis (MVPA): Random permutations and cluster
+       size control. NeuroImage, 65, 69--82.
+    .. [2] Smyth, G. K., & Phipson, B. (2010). Permutation P-values Should
+       Never Be Zero: Calculating Exact P-values When Permutations Are
+       Randomly Drawn. Statistical Applications in Genetics and Molecular
+       Biology, 9, 1--12.
```
yarikoptic: time will come.... https://github.com/yarikoptic/duecredit

mvpa2/algorithms/group_clusterthr.py
```diff
@@ -0,0 +1,427 @@
...
+        #
+        # Step 0: bootstrap maps by drawing one for each chunk and average them
+        # (do N iterations)
+        # this could take a lot of memory, hence instead of computing the maps
+        # we compute the source maps they can be computed from and then (re)build
+        # the matrix of bootstrapped maps either row-wise or column-wise (as
+        # needed) to save memory by a factor of (close to) `n_bootstrap`
+        # which samples belong to which chunk
+        chunk_samples = dict([(c, np.where(ds.sa[chunk_attr].value == c)[0])
+                                    for c in ds.sa[chunk_attr].unique])
```
yarikoptic: just a nitpicking -- (although iirc might be a part of PEP8). it is easier to read whenever there is not excessive indentation and next line indented at the same level of the code block -- here right after ```dict([```

mvpa2/algorithms/group_clusterthr.py
```diff
@@ -0,0 +1,427 @@
...
+        # Step 2: threshold all NULL maps and build distribution of NULL cluster
+        #         sizes
+        #
+        cluster_sizes = Counter()
+        # recompute the bootstrap average maps to threshold them and determine
+        # cluster sizes
+        if 'mapper' in ds.a:
+            dsa = dict(mapper=ds.a.mapper)
+        else:
+            dsa = {}
```
yarikoptic: more pythonic construct would be

```
dsa = dict(mapper=ds.a.mapper) if 'mapper' in ds.a else {}
```

mvpa2/algorithms/group_clusterthr.py
```diff
@@ -0,0 +1,427 @@
...
+        if __debug__:
+            debug('GCTHR', 'Estimate NULL distribution of cluster sizes')
```
yarikoptic: nitpick.  IMHO imperative tone is a bit misleading here -- "Estimating ..." would be more representative and that is what we usually use in our debug msgs before something is going to happen

mvpa2/tests/test_mapper.py
```diff
@@ -457,3 +457,15 @@ def test_addaxis():
...
     ds4 = AddAxisMapper(pos=4)(ds)
     assert_array_equal(ds4.shape, ds.shape + (1, 1))
+
+def test_identity_mapper():
+    idm = IdentityMapper()
+    # doesn't matter what you throw at it
+    for s in ('i_am_the_test',
+              range(5),
+              np.arange(12).reshape(4,3),
+              Dataset(np.arange(12).reshape(4,3))):
```
yarikoptic: you could have used ```@sweep_args```, which would be handy if e.g. only one of them fails -- all the others still get tested etc

mvpa2/tests/test_group_clusterthr.py
```diff
@@ -0,0 +1,241 @@
...
+    #
+    # check output
+    #
+    # samples unchanged
+    assert_array_equal(blob.samples, res.samples)
+    # need to find the big cluster
+    assert_true(len(res.a.cluster_probs_uncorrected) > 0)
+    assert_equal(len(res.a.cluster_probs_uncorrected), res.fa.clusters_featurewise_thresh.max())
+    # probs need to decrease with size
+    assert_true(res.a.cluster_probs_uncorrected[1] >= res.a.cluster_probs_uncorrected[2])
```
yarikoptic: there is ```nose.tools.assert_greater_equal``` which would make error msg more informative in case of failed test

dinga92:
If the surface based data can be represented as 2D then it should already work now, cluster creation/counting works for 2D and 3D data the same, however it's not tested. Different rules for the voxel connection should be almost trivial to add some more complex neighborhood functions can be a bit more tricky

nno:
@dinga92 even though surface is presented in a dataset structure as 2D (samples x nodes), the neighbourhood information is based on other information about the surface (for direct-neighbor surface-based clustering; the topology is sufficient to define the neighbourhood information). It may be preferable to have a standard solution (e.g. neighbourhoods) for different types of datasets.

mvpa2/algorithms/group_clusterthr.py
hanke: coming

mvpa2/algorithms/group_clusterthr.py
hanke: With 10000 subjects even 1 permutation per subject should be reasonable ;-)

mvpa2/algorithms/group_clusterthr.py
hanke: You can find examples in this code file where this kind of suggested indentation would be incomprehensible (nested list comprehensions). Unless you force me, I'd like to keep it this way.

mvpa2/algorithms/group_clusterthr.py
hanke: Good point...coming.

mvpa2/algorithms/group_clusterthr.py
hanke: Coming.

mvpa2/tests/test_mapper.py
hanke: Indeed... coming.

mvpa2/tests/test_group_clusterthr.py
hanke: But it cannot deal with arrays.

hanke:
BTW: Finished first full analysis on a dataset with 20 subjects, 100 permutations each, little more than 0.5 million features, 100k bootstrap samples. Purely single-threaded it took about 18 hours to compute. Virtually all time was spent on training/estimating the two NULL distributions. Results look sensible at first glance.

Getting the 20x100 NULL maps on the other hand took about 4 CPU years (1.2mm resolution), dense searchlight (Thanks $DEITY for a compute cluster!). So even if there is a lot of potential to speed this code up, the cluster-thresholding takes no time in comparison to the first step.

Doing tests on a more conventional dataset (~50k features) next...

hanke:
@yarikoptic any idea what travis wants to say with:

````
  mvpa2.tests.test_group_clusterthr.test_cluster_count ... /home/travis/build.sh: line 41: 22288 Segmentation fault      /usr/bin/nosetests -v mvpa2 --with-doctest -s
````

File: mvpa2/algorithms/group_clusterthr.py [hunk 10]
```diff
@@ -0,0 +1,424 @@
...
+    """Statistical evaluation of group-level average accuracy maps
+
+    This algorithm can be used to perform cluster-thresholding of
+    searchlight-based group analyses. It implements a two-stage procedure that
+    uses the results of within-subject permutation analyses, estimates a per
+    feature cluster forming threshold (via bootstrap), and uses the thresholded
+    bootstrap samples to estimate the distribution of cluster sizes in
+    group-average accuracy maps under the NULL hypothesis, as described in [1]_.
+
+    Note: this class implements a modified version of that algorithm. The
```
dinga92: One more difference: the original paper uses benjamini's step down fdr correction. This exact algorithm is not available in statsmodels. Probably it's not a big difference but we should state it

yarikoptic:
Not sure yet beyond what it already says: that test causes it to segfault. First I would have checked it on a similar precise chroot, and if not reproduced, troubleshoot straight on github ;-) I can check tomorrow may be whenever back in work seat

yarikoptic:
ok -- segfault replicates in a precise local environment (I would recommend anyone to have it handy for troubleshooting similar problems with travis + use schroot to enter it). well -- smells like a scipy bug which obviously was fixed since then:

```
(gdb) bt 5
#0  0x00007ffff2fddf43 in NI_Label (input=0x26138c0, strct=0x2613940, max_label=0x7fffffff78b8, output=
    0x2613a40) at scipy/ndimage/src/ni_measure.c:137
#1  0x00007ffff2fcecc7 in Py_Label (obj=0x0, args=
    (<numpy.ndarray at remote 0x26138c0>, <numpy.ndarray at remote 0x2613940>, <numpy.ndarray at remote 0x2613a40>)) at scipy/ndimage/src/nd_image.c:693
#2  0x0000000000486896 in PyCFunction_Call (func=<built-in function label>, arg=
    (<numpy.ndarray at remote 0x26138c0>, <numpy.ndarray at remote 0x2613940>, <numpy.ndarray at remote 0x2613a40>), kw=0x0) at ../Objects/methodobject.c:81
#3  0x000000000052645a in call_function (pp_stack=0x7fffffff7a10, oparg=3) at ../Python/ceval.c:4021
#4  0x0000000000521326 in PyEval_EvalFrameEx (f=
    Frame 0x2a5ec90, for file /usr/lib/python2.7/dist-packages/scipy/ndimage/measurements.py, line 159, in label (input=<numpy.ndarray at remote 0x26138c0>, structure=<numpy.ndarray at remote 0x2613940>, output=<numpy.ndarray at remote 0x2613a40>, ii=3, return_value=<numpy.ndarray at remote 0x2613a40>), throwflag=0)
    at ../Python/ceval.c:2666
(More stack frames follow...)
```

```
(gdb) py-list 
 154            if output.dtype.type != numpy.int32:
 155                raise RuntimeError('output type must be int32')
 156        else:
 157            output = numpy.int32
 158        output, return_value = _ni_support._get_output(output, input)
>159        max_label = _nd_image.label(input, structure, output)
 160        if return_value is None:
 161            return max_label
 162        else:
 163            return return_value, max_label
 164    
```

So, not sure what we could do about it really besides implementing 'label' functionality in Python for such cases of older scipy

hanke:
@dinga92 @yarikoptic @nno Now as parallel as you want. Thanks to @miykael

hanke:
@yarikoptic thanks for looking into the segfault. Tests still need to improve, maybe we can stop triggering this bug without loosing the test entirely.

hanke:
@miykael Open parallelization issues: Currently uses the "threading" backend, i.e. largely single CPU performance. Looking into memmapping to use the multiprocessing backend instead.

nno:
@hanke @miykael @yarikoptic it seems that parallelisation uses another library (joblib ) than what we use for the searchlight (pprocess). This reminds me of some earlier efforts I did to provide a uniform parallalization interface [https://github.com/nno/PyMVPA/tree/_tent/parallel]. (Note that in that branch I tried to parallalize RepeatedMeasure as well, but this attempt failed because conditional attributes were difficult to handle).

Would it be worthwhile to try to revisit this issue and see if a we can come up with a uniform parallelization solution?

hanke:
@nno definitely worthwhile. I'd prefer to move away from pprocess and go towards a maintained solution. yarik had at least one issue with joblib in the past https://github.com/joblib/joblib/issues/76
so it may turn out to be suboptimal. However, with respect to this PR the issue is slightly different, as this is pretty much a standalone algorithm. Whereas RepeatedMeasure and Searchlight tend to be nested.

In any case I need to get deeper into the topic. Trying this implementation now on a large dataset, and I cannot use more than two processes due to memory constraints.

nno:
@hanke Fine with me to keep it separate for this PR. Just to make sure we don't forget, I added this idea to the issue tracker (#287).

hanke:
Sidenote: In order to use this code with largish datasets (>10GB in my case) I need to do something like the following before I call train() with the dataset that has the permutations `perm_ds`:

````
import tempfile
from joblib import dump, load
temp_folder = tempfile.mkdtemp()
filename = os.path.join(temp_folder, 'dogrpstat.mmap')
if os.path.exists(filename): os.unlink(filename)
_ = dump(perm_ds.samples, filename)
perm_ds.samples = load(filename, mmap_mode='r+')
import gc
_ = gc.collect()
````

This writes the sample array to disk and memory maps its, dropping the resident memory size of the process (A LOT) before joblib starts forking children. Starting with a lower demand for the parent process leaves more room for children.

May be worth providing a dataset function to let a dataset memory map itself... @yarikoptic ?

**Comment by hanke** [comment]
@nno @dinga92 @yarikoptic @miykael I think this is ready to merge. Please have at it.

File: mvpa2/algorithms/group_clusterthr.py (hunk 3 above)
Comment (yarikoptic) [inline mvpa2/algorithms/group_clusterthr.py]: i don't think so  :-) FWIW

File: mvpa2/tests/test_group_clusterthr.py [hunk 11]
```diff
@@ -0,0 +1,303 @@
...
+        assert_array_equal(M, ds)
+        expected_result = Counter(np.hstack([gct._get_map_cluster_sizes(
+                                         thresholded_cluster_sizes_map)]*10))
+        th_map = np.ones(cluster_sizes_map.flatten().shape) * thr
+        # threshold dataset by hand
+        ds.samples = ds.samples > th_map
+        assert_array_equal(expected_result,
+                           gct.get_cluster_sizes(ds))
+
+        #dumm_null_dist = range(10)
```
yarikoptic: what is up with all this commented out testing?  is it wrong, or ... ? should be prune or some comment at least left why commented out?

mvpa2/tests/test_group_clusterthr.py
```diff
@@ -0,0 +1,303 @@
...
+
+def test_cluster_count():
+    # we get a ZERO cluster count of one if there are no clusters at all
+    # this is needed to keept track of the number of bootstrap samples that yield
+    # no cluster at all (high treshold) in order to compute p-values when there is no
+    # actual cluster size histogram
+    assert_equal(gct._get_map_cluster_sizes([0,0,0,0]), [0])
+    # if there is at least one cluster: no ZERO count
+    assert_equal(gct._get_map_cluster_sizes([0,0,1,0]), [1])
+    for i in range(2):  # rerun tests for bool type of test_M
```
yarikoptic: I would recommend to make it a "parametric" test where  `i` is an argument.  Two possible ways

1.  use our sweepargs```@sweepargs(i=[1,2])```

2. making it "nose parametric":
 a. rename this one to e.g. _test_cluster_count(i)
 b. make
```python
def test_cluster_count():
       for i in [1,2]:
           yield _test_cluster_count, i
```

imho we should start using 2 more often ;)

yarikoptic:
just for the record -- to merge it, we should deal with failing test one way (somehow fix) or another (skip on that elderly scipy, ideally only the failing piece)

hanke:
Ok, that worked. One test disabled for elderly SciPys -- still leaving plenty of calls to the label() function even on those systems. Ready to merge.

Will revisit this page later today and push the button, if nobody else does it before.

PR #286, opened 2015-04-20
ENH: Prevent potentially excessize memory use by default.

hanke:
Please merge this. Forcing the inclusion of the GLM model (as we do now) makes this unusable for the forrets gump data -- I need some 70+ GB for a simple 3 regressor fit. With this change, I can do it with 68GB less ;-)

### DIFF
--- a/mvpa2/datasets/eventrelated.py
+++ b/mvpa2/datasets/eventrelated.py
@@ -385,7 +385,7 @@ def extract_boxcar_event_samples(
 
 def fit_event_hrf_model(
         ds, events, time_attr, condition_attr='targets', design_kwargs=None,
-        glmfit_kwargs=None, regr_attrs=None):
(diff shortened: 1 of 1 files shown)


yarikoptic:
could you please Travis?

hanke:
@yarikoptic a happy camper now...

yarikoptic:
would you happen to please coveralls as well? ;)  e.g. by @sweepargs for providing or not the model in existing test?

hanke:
@yarikoptic not sure what you are asking for. Running the test twice to get or not get an element in a dict that we know very little of and do nothing with? Maybe I am not getting that report, but it seem like it has very little to do with the issue at hand.

yarikoptic:
I just reacted to decreased coverage... ok -- whatever -- merging ;-)

hanke:
Ah, cosmetics ;-) I'll make up for those 0.01% in another PR -- promised!

Issue #264, opened 2014-12-31
verify assumptions and adjust tests for nibabel 2.0.0 


yarikoptic:
```
======================================================================
FAIL: mvpa2.tests.test_niftidataset.test_assumptions_on_nibabel_behavior
----------------------------------------------------------------------
Traceback (most recent call last):
  File "c:\buildslaves\win7-32-mini\pymvpa-py2_7-win32\build\venv\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "c:\buildslaves\win7-32-mini\pymvpa-py2_7-win32\build\venv\lib\site-packages\mvpa2\testing\tools.py", line 165, in newfunc
    func(*(arg + (filename,)), **kw)
  File "c:\buildslaves\win7-32-mini\pymvpa-py2_7-win32\build\venv\lib\site-packages\mvpa2\tests\test_niftidataset.py", line 474, in test_assumptions_on_nibabel_behavior
    assert_equal(slope, 1.0)
AssertionError: None != 1.0
```
http://nipy.bic.berkeley.edu/builders/pymvpa-py2.7-win32/builds/62/steps/shell_4/logs/stdio

hanke:
To me this seems like an inappropriate test. Nibabel says:

````
slope : None or float
   scaling (slope).  None if there is no valid scaling from these fields
inter : None or float
   offset (intercept). None if there is no valid scaling or if offset is
   not finite.
````

NIfTI standard says: If there is no scaling set, don't scale. Consequently, despite the change in header setting, there is no change in behavior.

@yarikoptic Agree that adjusting the test is the way forward?

yarikoptic:
sure -- go ahead

PR #288, opened 2015-05-04
BF: Make nibabel tests more robust (fixes gh-264)

hanke:


hanke:
Merging this now. A PR that increases the coverage will appear later today.

PR #290, opened 2015-05-04
MSC: update copyright and developer info to reflect current state

nno:
Very minor update for copyright year (2014 -> 2015)  and affiliation. 

@hanke, @yarikoptic, are there new contributors we should add?


hanke:
@nno I was trying to keep the contributors list up-to-date -- I think we are fine for now.

PR #295, opened 2015-05-04
RF: clean split of dataset generators vs. loaders (plus some general code love)

hanke:


hanke:
Please wait with a potential merge for the comming coverage boosting commit.

mvpa2/datasets/sources/native.py
```diff
@@ -0,0 +1,110 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the PyMVPA package for the
+#   copyright and license terms.
+#
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""Loaders for PyMVPA's own demo datasets"""
```
yarikoptic: may be make it "pymvpa_demo.py" instead of "native.py" to be more explicit/less confusing?

hanke:
@nno @yarikoptic The interesting commit is in. Please take a look.

mvpa2/datasets/sources/native.py
hanke: @yarikoptic Yeah, I was debating that with myself. pymvpa_demo is not good: sure it is pymvpa, all of this is. demo alone is not good, anything could be a demo. Hence, I settled with native. These are dataset source, this is our native one.

I was quite pleased with myself to come up with this reasoning.

yarikoptic: ;-) oh well -- I don't want to diminish your self-esteem  so let it be "native"

hanke:
The day of the short fuse: merging...

PR #296, opened 2015-05-04
ENH(TST)+BF:  rudimentary regression testing for stored datasets

yarikoptic:
I am yet to upload that load somewhere where it would become available (having difficulty with "annex copy --to=alioth"), so test should get skipped on travis and not to cause failure


yarikoptic:
0 -- is done.  Ok -- now we have a clone on pymvpa.org and I have pushed git-annex branch for now only into my clone, but this should work

```
git clone git://github.com/yarikoptic/PyMVPA -b enh/regress-hdf5
cd PyMVPA
git annex get .
```

hanke:
@nno I am in favor of merging this. It is useful infrastructure and tests keep running smoothly without the required annex. What do you think?

PR #297, opened 2015-05-05
All equal broken

hanke:
This is just to make the bug obvious. Fix needs to come.


hanke:
@nno please have a look at this one. This fix is ugly, but it seems to work.

hanke:
Ready to merge from my POV.

nno:
@hanke looks good to me, thanks.

PR #298, opened 2015-05-05
Become more independent from nibabel's internals

hanke:
This is step one: no longer store nibabel internals in datasets.

NOTE: not for merge -- let's see what the test machinery says...

This causes issues with nibabel API changes over time. The next solution
is, instea dof storing an instance of the image header, to store

imgtype <string>
imgaffine <arrray>
imhdr dict(array)

These dtypes should yield a much more compact output when serialized to
HDF5 and are sufficient to achieve all of the previous functionality
without compromise.

This commit implements the technical changes and adjusts the tests, as
far as I could see.

(fingers crossed)

### DIFF
--- a/mvpa2/base/dataset.py
+++ b/mvpa2/base/dataset.py
@@ -795,7 +795,18 @@ def all_equal(x, y):
     try:
         eq = x == y
     except:
-        return False
+        # we can get here, in case of dictionaries with non-simple values
+        # (e.g. arrays)
+        try:
+            if not set(x.keys()) == set(y.keys()):
+                return False
+            x = [x[k] for k in x]
+            y = [y[k] for k in y]
+            # this is just for fooling the next test
+            eq = [0, 1]
+        except:
+            # cannot do more than try
+            return False
 
     # eq could be a numpy array or similar. See if it has a length
     try:
--- a/mvpa2/datasets/mri.py
+++ b/mvpa2/datasets/mri.py
@@ -20,27 +20,23 @@ __docformat__ = 'restructuredtext'
 from mvpa2.base import externals
 externals.exists('nibabel', raise_=True)
 
-import sys
 import numpy as np
-from mvpa2.support.copy import deepcopy
-from mvpa2.misc.support import Event
-from mvpa2.base.collections import DatasetAttribute
 from mvpa2.base.dataset import _expand_attribute
 
 if __debug__:
     from mvpa2.base import debug
 
 from mvpa2.datasets.base import Dataset
-from mvpa2.mappers.fx import _uniquemerge2literal
 from mvpa2.mappers.flatten import FlattenMapper
-from mvpa2.mappers.boxcar import BoxcarMapper
 from mvpa2.base import warning
+
 
 def _hdr2dict(hdr):
     """Helper to convert a NiBabel image header to a dict of arrays"""
     kv = dict(hdr)
     kv['hdrtype'] = hdr.__class__.__name__
     return kv
+
 
 def _dict2hdr(kv):
     """Counterpart of ``_hdr2dict()``"""
@@ -159,7 +155,7 @@ def map2nifti(dataset, data=None, imghdr=None, imgtype=None):
     # with imghdr.get_data_dtype()
 
     if issubclass(imgtype, nibabel.spatialimages.SpatialImage) \
-       and (imghdr is None or hasattr(imghdr, 'get_data_dtype')):
+            and (imghdr is None or hasattr(imghdr, 'get_data_dtype')):
         # we can handle the desired image type and hdr with nibabel
         # use of `None` for the affine should cause to pull it from
         # the header
@@ -287,7 +283,8 @@ def fmri_dataset(samples, targets=None, chunks=None, mask=None,
         ds.a['imghdr'] = _hdr2dict(imghdr)
     except:
         if __debug__:
-            debug('DS_NIFTI', 'Failed to store header info as attribute (src: %s)' \
+            debug('DS_NIFTI',
+                  'Failed to store header info as attribute (src: %s)'
                   % (imghdr.__class__,))
     # If there is a space assigned , store the extent of that space
     if sprefix is not None:
@@ -297,8 +294,8 @@ def fmri_dataset(samples, targets=None, chunks=None, mask=None,
         # TODO extend with the unit
     if tprefix is not None:
         ds.sa[tprefix + '_indices'] = np.arange(len(ds), dtype='int')
-        ds.sa[tprefix + '_coords'] = np.arange(len(ds), dtype='float') \
-                                     * _get_dt(imghdr)
+        ds.sa[tprefix + '_coords'] = \
+            np.arange(len(ds), dtype='float') * _get_dt(imghdr)
         # TODO extend with the unit
 
     return ds
@@ -355,7 +352,7 @@ def _load_anyimg(src, ensure=False, enforce_dim=None):
       If there is a problem with data (variable dimensionality) or
       failed to load data and ensure=True.
     """
-    imgdata = imghdr = imgtype = None
+    imgdata = imghdr = None
 
     # figure out whether we have a list of things to load and handle that
     # first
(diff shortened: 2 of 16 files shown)


hanke:
Test failure comes from missing fix from gh-297. Merging it in.

hanke:
NOTE: This new code will strip any NIfTI extensions during a conversion cycle from NIfTI->ds->NIfTI. I don't think this is a problem, but it is worth mentioning.

Actually, we could consider stripping even more information, but I felt going with the plain NIfTI header is OK.

hanke:
@nno @yarikoptic Seems ready to merge. Please inspect.

mvpa2/datasets/mri.py
```diff
@@ -270,8 +283,18 @@ def fmri_dataset(samples, targets=None, chunks=None, mask=None,
             ds.fa[fattr] = ds.a.mapper.forward1(value)
 
     # store interesting props in the dataset
-    ds.a['imghdr'] = imghdr
-    ds.a['imgtype'] = imgtype
+    ds.a['imgaffine'] = img.get_affine()
```
nno: how about using get_best_affine() instead? According to the documentation, this "Select[s] best of available transforms"

mvpa2/datasets/mri.py
```diff
@@ -270,8 +283,18 @@ def fmri_dataset(samples, targets=None, chunks=None, mask=None,
...
+    ds.a['imgtype'] = img.__class__.__name__
```
nno: Getting the image type and affine matrix seems duplicated in the convert_dataset function in the new file tools/strip_nibabel. Can we put this functionality in one place?

mvpa2/datasets/mri.py
hanke: get_best_affine() is a header function. Images only have get_affine(). It should do the right thing.
FYI: Not all image headers have affines. In the MINC case the image is the only object that has it.

mvpa2/datasets/mri.py
hanke: We could, but it is just three lines. The tools was intented to be as agnostic as possible wrt versions of installed packages.

mvpa2/base/dataset.py
```diff
@@ -795,7 +795,18 @@ def all_equal(x, y):
...
         eq = x == y
     except:
-        return False
+        # we can get here, in case of dictionaries with non-simple values
+        # (e.g. arrays)
+        try:
+            if not set(x.keys()) == set(y.keys()):
+                return False
+            x = [x[k] for k in x]
+            y = [y[k] for k in y]
```
nno: It seems that "k in x" can return keys in arbitrary order, so maybe better to do:

keys=list(x.keys()); % by python3 compatible
x = [x[k] for k in keys]
y = [y[k] for k in keys]

yarikoptic:
yet to review...  But what about getting #296  in first? 
 I  will make those annexed files available today one way or another

hanke:
@yarikoptic i don't mind, but it is failing right now.

hanke:
@yarikoptic you to be clear, the purpose of this one is to decouple from NiBabel completely. Unless you want to export to Nifti again, you don't need it once a dataset is created.

hanke:
@yarikoptic Looked at gh-296. Works for me. But the premise seems that we need matching nibabel versions. However, with this PR this should no be a requirement anymore. Hence if we want to order merges (don't see why), this should go in first and gh-296 afterwards in order to be able to exercise the new flexibility.

tools/strip_nibabel
```diff
@@ -0,0 +1,73 @@
...
+Get the NiBabel repository from github and checkout the version you need::
+
+  % git clone  https://github.com/nipy/nibabel /tmp/nibabel
+  % cd /tmp/nibabel
+  % git checkout 1.1.0
+
+Now call this script with the path to your old NiBabel, the HDF5 file with
+the to-be-converted datasets, and the output filename.
+
+  % tools/strip_nibabel /tmp/nibabel olddata.hdf5 newdata.hdf5
```
yarikoptic: I would recommend (an alternative path) using virtualenv, although would need to be installed, smth like

```shell
virtualenv --system-site-packages venv-nibabel1.0
venv-nibabel1.0/bin/pip install nibabel==1.1.0
venv-nibabel1.0/bin/python tools/strip_nibabel /tmp/nibabel olddata.hdf5 newdata.hdf5
```

tools/strip_nibabel
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/python
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the PyMVPA package for the
+#   copyright and license terms.
+#
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""Convert NiBabel objects in PyMVPA datasets attributes into a portable form
```
yarikoptic: great idea btw, but why not PyMVPA command???

tools/strip_nibabel
yarikoptic: theoretically, we could even automate all of that within the command -- folks would just specify the nibabel version and we git clone (or virtualenv) and run this using that nibabel?  should be very doable and not too fragile

tools/strip_nibabel
yarikoptic: btw by placing it under mvpa2/ codebase we at least would have better opportunity to provide a unittest for this beastie. Especially if that PR for testing compatibility would get merged -- i.e. if we get our test files for this conversion to be tested

tools/strip_nibabel
hanke: Seems overkill for a one time effort. I have no experience with virtualenv, but feel free to tweak. From my POV I have no preference for either git or virtualenv -- unless virtualenv would pull in tons of 3rd-party packages before it could do the job. The currently implemented approach takes seconds to be functional.

tools/strip_nibabel
hanke: I considered it a one time helper. Current PyMVPA can ready old datasets transparently anyway, and write new-style ones automatically.

mvpa2/base/dataset.py
yarikoptic: and not to forget ```sorted(x.keys())```

hanke: @yarikoptic  Are you looking at the latest code?

hanke: @yarikoptic Sorry, incomplete message. The PR that actually had this change and got merged has an improved version without this issue.

tools/strip_nibabel
yarikoptic: how do we know that it works? ;-)
and not for me to tell you tales about "one time helpers" getting their own lives ;)

Let me finish this one up by
0. addressing sorted dict issue in comparison
1. push those test datasets somewhere accessible (I am thinking about having a git/annex clone under pymvpa.org/_files)
2. RF this one into a PyMVPA command with some minimal tests

Deal?

hanke: 0. Has been dealt with in the actual PR (merged)
1. Thanks
2. Thanks

yarikoptic:
ok -- dealt with test datasets   :  https://github.com/PyMVPA/PyMVPA/pull/296#issuecomment-99094778   I will try now the "2"

yarikoptic:
BTW -- I wondered that may be we should also store in `ds.a.external_versions` some information about versions of externals (similarly to how I did in those regression datasets)?   That could simplify migrations later on...?

hanke:
@yarikoptic We will never success in turning this serialized madness into a proper container. Don't think we should even make the impression of an attempt.

yarikoptic:
;-)  but I don't see why we shouldn't safeguard our stepping on our own toes since we are providing this functionality

hanke:
There is no immediate benefit. Even if we know that versions we cannot do much in most cases. Moreover, storing a dict is actually pretty demanding ...

hanke:
@yarikoptic gave his blessing on another channel. Merging.

hanke:
By hand....later...

Issue #301, opened 2015-05-05
Failing tests with assert_objectarray_equal


nno:
Currently OSX 10.5, 10.6, 10.8 [e.g. http://nipy.bic.berkeley.edu/builders/pymvpa-py2.7-osx-10.6/builds/276/steps/shell_5/logs/stdio] may give errors in assert_objectarray_equal, which seems to be caused by rounding errors.

@yarikoptic @hanke would it be good to add another function, assert_objectarray_almost_equal, to deal with such situations?

FAIL: mvpa2.tests.test_fxmapper.test_fx_native_calls
 ----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/buildslave/osx-10.6/pymvpa-py2_7-osx-10_6/build/venv/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/buildslave/osx-10.6/pymvpa-py2_7-osx-10_6/build/venv/lib/python2.7/site-packages/nose/util.py", line 620, in newfunc
    return func(*arg, **kw)
  File "/Users/buildslave/osx-10.6/pymvpa-py2_7-osx-10_6/build/venv/lib/python2.7/site-packages/mvpa2/tests/test_fxmapper.py", line 220, in test_fx_native_calls
    assert_objectarray_equal(dsm1.samples, dsm2.samples)
  File "/Users/buildslave/osx-10.6/pymvpa-py2_7-osx-10_6/build/venv/lib/python2.7/site-packages/mvpa2/testing/tools.py", line 326, in assert_objectarray_equal
    assert_array_equal(x, y)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/numpy/testing/utils.py", line 739, in assert_array_equal
    verbose=verbose, header='Arrays are not equal')
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/numpy/testing/utils.py", line 665, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
 Single scenario lead to failures of unittest test_fx_native_calls:
  on
    f=average :
     
Arrays are not equal

(mismatch 66.6666666667%)
 x: array([[ 0.343741, -0.016952,  0.013869, -0.004369, -0.016687,  0.312285]])
 y: array([[ 0.343741, -0.016952,  0.013869, -0.004369, -0.016687,  0.312285]])
    f=mean :

[...]

PR #303, opened 2015-05-05
BF+TST: do not use objectarray_equal for numerical comparisons

nno:
Addresses #301


hanke:
+1 in general. Let's wait for travis.

Issue #302, opened 2015-05-05
Failing tests in test_stats_sp.test_tsboxplot


nno:
On some mac boxes [see http://nipy.bic.berkeley.edu/builders/pymvpa-py2.7-osx-10.6/builds/276/steps/shell_5/logs/stdio] two tests fail.

As in #229, tests on NeuroDebian fail with recent numpy/scipy but not with older versions:
fail: numpy 1.10.0.dev0+63e8bab and scipy 0.16.0.dev0+1306448
pass: numpy 1.6.2 and scipy 0.10.0.

(Note: numpy is reported as an "older" version even though I pulled straight from git://github.com/numpy/numpy/git and then compiled & installed)

@yarikoptic @hanke For now I added this to the Release 2.4 milestone, I hope you agree?
```
FAIL: mvpa2.tests.test_stats.test_tsboxplot
 ----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/buildslave/osx-10.6/pymvpa-py2_7-osx-10_6/build/venv/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/buildslave/osx-10.6/pymvpa-py2_7-osx-10_6/build/venv/lib/python2.7/site-packages/mvpa2/tests/test_stats.py", line 175, in test_tsboxplot
    assert_equal(outlier.mask.sum(), np.prod(outlier.shape) - 1)
AssertionError: 18 != 19

 ======================================================================
FAIL: mvpa2.tests.test_stats_sp.test_tsboxplot
 ----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/buildslave/osx-10.6/pymvpa-py2_7-osx-10_6/build/venv/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/buildslave/osx-10.6/pymvpa-py2_7-osx-10_6/build/venv/lib/python2.7/site-packages/nose/util.py", line 620, in newfunc
    return func(*arg, **kw)
  File "/Users/buildslave/osx-10.6/pymvpa-py2_7-osx-10_6/build/venv/lib/python2.7/site-packages/mvpa2/tests/test_stats.py", line 175, in test_tsboxplot
    assert_equal(outlier.mask.sum(), np.prod(outlier.shape) - 1)
AssertionError: 18 != 19
```

yarikoptic:
[...]

Issue #229, opened 2014-06-30
fresh cosmo_* tests failures under wheezy/sparc


yarikoptic:
Please check it out at
http://nipy.bic.berkeley.edu/builders/pymvpa-py2.7-wheezy-sparc/builds/14/steps/shell_6/logs/stdio
If those are sparc specific (ie do not reproduce in neurodebian wheezy vm) -- I will give you access to that beast

nno:
This test now fails on my machine (OSX 10.9.5) with numpy 1.9.0 and scipy 0.14.0.
On NeuroDebian it runs fine with numpy 1.6.2 and scipy 0.10.0.
I'll try to figure out which of these components is causing it and try to fix it.

nno:
On NeuroDebian with the latest numpy (1.10.0.dev0+63e8bab) and scipy (0.16.0.dev0+1306448) the tests fail as well. The following code pins down the issue:
```
import numpy as np
arr=np.array

x=arr([[1]])
y=arr([[1,3]])

xy=arr([[x,y]], dtype='O')
```

which crashes with recent numpy and scipy:
```
Traceback (most recent call last):
  File "test.py", line 7, in <module>
    xy=arr([[x,y]], dtype='O')
ValueError: could not broadcast input array from shape (2) into shape (1)
```

nno:
This issue should be addressed by PR #299.

yarikoptic:
#299 was merged, thus closing

PR #304, opened 2015-05-05
deal properly with NaN values in compute_ts_boxplot_stats

nno:
Attempts to address #302


hanke:
@nno Looks good

nno:
As in #229, there is different behaviour depending on how recent numpy/scipy is, and now the tests fail for older versions. I'll work on a fix.

yarikoptic:
I guess this one can get in now, @nno?

nno:
@yarikoptic yes I think it is ready, I will merge.

PR #307, opened 2015-05-06
WiP: Kitchen sink of updates for v2.4

hanke:
Low-hanging fruit first.


hanke:
As of 2125636f4fa1ffccb8f99e1e63d58587464308e5 all examples pass again (on my machine).

hanke:
@yarikoptic Sphinx extension update was prompted by autodoc failure that is also present in the master.

however, I am still having issues with generating the full modref.

hanke:
@nno @yarikoptic I am giving up for tonight. It is probably a simple issue -- that I cannot see. My guess is in the last commit message.

yarikoptic:
I will have a look. Cheers!

yarikoptic:
if only I knew what you meant about problem "with generating the full modref."  After I have managed to install all our optional depends it generated... BUT as before (apparently if you look at http://www.pymvpa.org/modref.html) it misses descriptions for the modules (which it had just fine: http://dev.pymvpa.org/modref.html).   I will assume that it is the problem and try to figure it out...?

yarikoptic:
ok -- seems to build fine with sphinx 1.3.1 from experimental... doing minor tune ups and will send your a PR later on

yarikoptic:
ah, silly me -- now I see what you meant , and found your BROKEN commit: https://github.com/hanke/PyMVPA/commit/3798ff21228ade69f1a046464b68840e56c3925a ... indeed -- reuploaded dev. built as is -- http://dev.pymvpa.org/modref.html#chap-modref  and indeed doesn't go into classes etc.  I will try now to check it out

yarikoptic:
uff -- sorry that I haven't found a proper/full solution but here is what I found in my debugging attempt:
not sure how it worked before and how it actually works for anyone now, since in autosummary/generate  it is the ```find_autosummary_in_lines``` function which supposed to find additional entries (Classes, Functions, etc) within generated module level .rst and cause generation of files for them.   It relies on having ```:toctree:``` and ```:template:``` being specified BUT module.rst template (picked up only from /usr/share/sphinx/ext/autosummary/templates/autosummary/module.rst if I see it correctly) has only "Rubric" to say what follows -- functions or classes or ... 
As a result, nothing gets picked up on the traversal of that module-level description.  I succesfully triggered generation of a class doc by adding those two for autosummary in module.rst:

```
   .. rubric:: Classes

   .. autosummary::
      :template: /usr/share/sphinx/ext/autosummary/templates/autosummary/class.rst
      :toctree:
```

but then -- what should be for functions, i.e. how did it work before?  there is no function.rst template (I see that sklearn folks have their own around, they also do not use as much of autosummary magic and manually set things up -- describe what will be the functions and what will be the...)

So -- as far as I see a possible resolution is to figure out if I have mistaken that module.rst (and other ```:template:```s get picked   by autosummary only from the system level), enhance module.rst with proper options for autosummary directives and it should work.

I am resting my case for today on this (just sent a minor PR to your repo along the way)

hanke:
@yarikoptic We do carry our own template in `doc/templates/autosummary`. Maybe they need to be tweaked, or it doesn't find them...

hanke:
Ha, I was right...

yarikoptic:
I do not have a good feel either this one is ready to merge or not ;) but please merge anything you consider worthwhile merging and I will run another snapshot package generation asap

hanke:
First half of the tutorial is updated, general functionality in place. Please merge ASAP.

Issue #245, opened 2014-09-21
Make code references to datadb go away


mih:
As per lunch beer discussion we need to move away from it and become more flexible and uniform regarding data discovery.

hanke:
It seems logical to switch this to a datalad based procedure, once we can rely on it. For know I don't know how to do it better than the status quo without duplicating work.

Moving to the next release.

PR #314, opened 2015-05-06
RF: Make the tutorial data less special.

hanke:
- load_datadb_tutorial_data() -> load_tutorial_data()

  we should not care where it is coming from

- Makefile TUT_DIR now points directly to the 'data' directory

- load_tutorial_data() can now also load the two other flavors (1slice
  and 25mm) that we ship with the sources. This should make it possible
  to (at least) run a lot of code that used to require the 300MB
  download before. It could also be used to speed tests up be some
  factor.

- this is a move into the general direction of gh-245


hanke:
Merging this is now a requirement for gh-307

mvpa2/datasets/sources/native.py
```diff
@@ -51,15 +52,18 @@ def load_example_fmri_dataset(name='1slice', literal=False):
     return ds
 
 
-def load_datadb_tutorial_data(path=os.path.join(
```
yarikoptic: couldn't we provide some deprecation path here?

hanke: like this:

````
load_datadb_tutorial_data = load_tutorial_data
````

?

hanke:
I am closing this one. It is merged into gh-307 and keep it open and separate is slowing things down....

PR #315, opened 2015-05-06
BF+NF+ENH: assert_datasets_equal (stub) + test_generate_testing_fmri_dataset

yarikoptic:
got broken due to recent renames

to fix first Travis failure had to make SmartVersion explicitly convert unicode in versions to str

all for good! ;)


yarikoptic:
just great -- so on older releases (h5py 2.0.1 I guess to be a curlpit) we can't save our dataset with ```dtype('<U5')```

nno:
@yoh looks good to me.

PR #320, opened 2015-05-18
ENH: Procrustean mapper now returns transpose as inverse

swaroopgj:
instead of computing inverse (as part of super class)
This should be faster and possibly avert any numerical errors that might creep up during numpy.linalg.inv()

### DIFF
--- a/mvpa2/mappers/procrustean.py
+++ b/mvpa2/mappers/procrustean.py
@@ -208,3 +208,10 @@ class ProcrusteanMapper(ProjectionMapper):
             d_r = np.linalg.norm(odatas[0] - res_r)/np.linalg.norm(odatas[0])
             debug('MAP_', "%s, residuals are forward: %g,"
                   " reverse: %g" % (repr(self), d_f, d_r))
+
+
+    def _compute_recon(self):
+        """For Procrustean mapper, inverse is transpose.
+        So, let's save on computation time and not introduce new erros(?)
+        """
+        return np.transpose(self._proj)
--- a/mvpa2/mappers/procrustean.py
+++ b/mvpa2/mappers/procrustean.py
@@ -212,6 +212,6 @@ class ProcrusteanMapper(ProjectionMapper):
 
     def _compute_recon(self):
         """For Procrustean mapper, inverse is transpose.
-        So, let's save on computation time and not introduce new erros(?)
+        So, let's skip computing inverse in the super class.
         """
         return np.transpose(self._proj)
--- a/mvpa2/mappers/procrustean.py
+++ b/mvpa2/mappers/procrustean.py
@@ -214,4 +214,9 @@ class ProcrusteanMapper(ProjectionMapper):
         """For Procrustean mapper, inverse is transpose.
         So, let's skip computing inverse in the super class.
         """
-        return np.transpose(self._proj)
+        # XXX Change pinv to superclass compute_recon?
+        if self.params.oblique:
+            #return ProjectionMapper._compute_recon(self)
+            return np.linalg.pinv(self._proj)
+        else:
+            return np.transpose(self._proj/self._scale**2) if self.params.scaling else np.transpose(self._proj)
--- a/mvpa2/tests/test_procrust.py
+++ b/mvpa2/tests/test_procrust.py
@@ -107,8 +107,8 @@ class ProcrusteanMapperTests(unittest.TestCase):
                 d_s_f_r = pm.reverse(d_s_f)
                 # Test if recon proj is true inverse except for high->low projection
                 if nf_s <= nf_t:
-                    self.assertTrue(norm(np.dot(pm._proj, pm._recon) - np.eye(pm._proj.shape[0])) < 1e-10,
-                                msg="Deviation from identity matrix is too large")
+                    assert_almost_equal(np.dot(pm._proj, pm._recon),np.eye(pm._proj.shape[0]),
+                                             err_msg="Deviation from identity matrix is too large")
                 dsfr = d_s_f_r - d_s
                 ndsfr = norm(dsfr)/norm(d_s)
                 if full_test:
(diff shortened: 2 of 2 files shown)


mvpa2/mappers/procrustean.py
```diff
@@ -208,3 +208,10 @@ def _train(self, source):
             d_r = np.linalg.norm(odatas[0] - res_r)/np.linalg.norm(odatas[0])
             debug('MAP_', "%s, residuals are forward: %g,"
                   " reverse: %g" % (repr(self), d_f, d_r))
+
+
+    def _compute_recon(self):
+        """For Procrustean mapper, inverse is transpose.
+        So, let's save on computation time and not introduce new erros(?)
```
nno: Spelling error in "erros", or was that intended?

swaroopgj: D'oh

mvpa2/mappers/procrustean.py
```diff
@@ -208,3 +208,10 @@ def _train(self, source):
             d_r = np.linalg.norm(odatas[0] - res_r)/np.linalg.norm(odatas[0])
             debug('MAP_', "%s, residuals are forward: %g,"
                   " reverse: %g" % (repr(self), d_f, d_r))
+
+
+    def _compute_recon(self):
+        """For Procrustean mapper, inverse is transpose.
```
yarikoptic: if it is -- please provide a unit-test which verifies that so we have some guarantee of such a promise

swaroopgj: There are two tests, one each in test_procrust.py and test_hyperalign.py, that check if the reconstruction is proper. Both would fail if inverse is not computed correctly. Should I write an additional test?

yarikoptic: re 1.  yes -- additional test which tests for particularly this, that whatever is returned from recon is the inverse (thus ```np.dot(_proj, _proj.T)``` and probably ```np.dot(_proj.T, _proj)``` close to identity ).  This way, if our other tests fail we would also know if that is solely due to this or not (**unit-** test)

2. ;-) unless we want to refactor it into some pure Procrustean and ObliqueProcrustean, which I don't think we should, this inversion here should be conditioned on either oblique was requested.  And then we should have tests for 1. in various scenarios

swaroopgj: Thanks for nudging me to write the test. I realized that I didn't handle scaling=True case.
Tests passed because they were testing correlation and/or zscoring data beforehand in case of real data.
Handled that and added a test now.

mvpa2/mappers/procrustean.py
```diff
@@ -208,3 +208,15 @@ def _train(self, source):
...
+        So, let's skip computing inverse in the super class.
+        """
+        # XXX Change pinv to superclass compute_recon?
+        if self.params.oblique:
+            #return ProjectionMapper._compute_recon(self)
+            return np.linalg.pinv(self._proj)
+        else:
+            return np.transpose(self._proj/self._scale**2) if self.params.scaling else np.transpose(self._proj)
```
yarikoptic: why is it ```_scale**2```?

swaroopgj: When scaling=True, proj is multiplied by uniform scaling=`scale` and the the matrix det is no longer 1.
So, reverse mapping matrix is still transpose but after multiplying with 1/scale. Here, I multiply with (1/scale) to reverse the scaling applied to proj first and then to apply reverse scaling for uniformly scaling by 1/scale. Hence, (1/scale**2)

yarikoptic: I understand the first but still lack the purpose of "reverse scaling for uniformly scaling by 1/scale" which imho shouldn't be applied -- are we applying scaling twice on the way "forward"?

swaroopgj: Well, to reverse map from target to source, we need to scale the target by `1/scale` for the same reason we multiply forward mapping matrix by scale in the main function. Does it make sense?
If not, run
r=get_random_rotation(2,2)*0.5
rpinv=np.linalg.pinv(r)
rtscaled=r.T/0.25
np.allclose(rpinv, rtscaled)

yarikoptic: I think I got it ;)  thanks

mvpa2/tests/test_procrust.py
```diff
@@ -105,7 +105,10 @@ def test_simple(self, svd, oblique):
 
                 # Test if we get back
                 d_s_f_r = pm.reverse(d_s_f)
-
+                # Test if recon proj is true inverse except for high->low projection
+                if nf_s <= nf_t:
+                    self.assertTrue(norm(np.dot(pm._proj, pm._recon) - np.eye(pm._proj.shape[0])) < 1e-10,
```
yarikoptic: ok -- make it kosher here (`assert_array_almost_equal` instead of norm etc) and we should be done ;)

swaroopgj: Done!

swaroopgj:
Done!

mvpa2/tests/test_procrust.py
```diff
@@ -105,7 +105,10 @@ def test_simple(self, svd, oblique):
 
                 # Test if we get back
                 d_s_f_r = pm.reverse(d_s_f)
-
+                # Test if recon proj is true inverse except for high->low projection
+                if nf_s <= nf_t:
```
yarikoptic: stop -- wouldn't we need similar conditioning inside ` _compute_recon` ?

swaroopgj: No need. Since pinv(R) is same as R.T (by definition), they both should give the same matrix, albeit pinv() going through more computations.

swaroopgj: Are you convinced?

PR #321, opened 2015-05-20
BF+ENH+DOC: An actually working motion QC plot command

hanke:
Not sure why the previous one was merged in a broken state. But I took the freedom to make it a bit more flexible.

### DIFF
--- a/mvpa2/cmdline/cmd_ofmotionqc.py
+++ b/mvpa2/cmdline/cmd_ofmotionqc.py
@@ -6,16 +6,43 @@
 #   copyright and license terms.
 #
 ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
-"""Generate a QC plot for BOLD fMRI motion estimates from an OpenFMRI dataset
+"""Generate a QC plot for BOLD fMRI motion estimates from an OpenFMRI dataset.
+
+The generated figure consists of two subplots: one for translation and one for
+rotation. The L2-norm for each motion type is plotted. Runs boundaries
+are indicated with dashed vertical lines. The following statistics are
+visualized
+
+1. Range across subjects (min, max) with a light gray shaded area
+
+2. 50% percentile around the mean with a medium gray shaded area
+
+3. Standard error of the mean (SEM; +/-) with a dark gray shaded area
+
+4. Median across subjects with a black line
+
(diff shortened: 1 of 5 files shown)


mvpa2/cmdline/cmd_ofmotionqc.py
```diff
@@ -53,63 +91,85 @@ def run(args):
...
-    segment_sizes = [len(d[0]) for d in data]
+    segment_sizes = [len(d[d.keys()[0]]) for d in data]
+
+    if not args.relative is None:
+        # recode per-subject estimates wrt their particular reference
+        ref = {subj: d[args.relative[1]]
+               for subj, d in data[args.relative[0]].iteritems()}
+        for d in data:
+            for subj in d:
+                if subj in d:
```
yarikoptic: not obvious to me: how can subj not be in d if we loop through it?

mvpa2/cmdline/cmd_ofmotionqc.py
```diff
@@ -53,63 +91,85 @@ def run(args):
     from mvpa2.datasets.sources.openfmri import OpenFMRIDataset
```
yarikoptic: could there be at least a smoke test for this now that we have a sample openfmri 1 subj tutorial_data?

mvpa2/cmdline/cmd_ofmotionqc.py
```diff
@@ -53,63 +91,85 @@ def run(args):
...
-    #pl.ylim((0,5))
-    pl.title('rotation')
-    pl.ylabel('estimate L2-norm in deg')
     pl.xlabel('time in fMRI volumes')
 
     if args.savefig is None:
         pl.show()
     else:
-        pl.savefig(args.safefig)
+        pl.savefig(args.savefig)
```
yarikoptic: unittest! unittest! ;)

mvpa2/cmdline/cmd_ofmotionqc.py
hanke: The moment we have a multi-subject demo dataset there will be a test.

hanke:
For the record: the test failure was a network issue.

hanke:
Merging this to have a working command. More beauty with next iteration.

Issue #312, opened 2015-05-06
Uniform abbrevation of os.path.join


hanke:
Just go through once and replace all with

````
from os.path import join as opj_
````

Vetos from @nno @yarikoptic ?

yarikoptic:
We having similar discussion in DataLad.... May be just join? :-) unlike ly to be confused

hanke:
My vote would be on `opj`. `join` is a string method that I'd rather not touch.

nno:
I would prefer ``pathjoin`` because that results in more readable code. Using ``_opj`` or ``obj`` would save 4 or 5 characters, but anyone somewhat new to the code would have to look op the import statement to see what it does.

hanke:
@nno I like that.

nno:
Addressed in #324.

PR #324, opened 2015-05-22
use "pathjoin" instead of "os.path.join"

nno:
Addresses #312. @hanke, @yarikoptic, any vetos?


yarikoptic:
Thanks @nno! Let's wait for travis and then merge if all kosher

nno:
Travis is happy, so I will go ahead and merge.

PR #325, opened 2015-05-26
Few small enhancements to the group cluster threshold code.

hanke:
Should be ready to go in, but may still get some changes during real-world testing.


hanke:
Thanks!

PR #328, opened 2015-06-10
Improved comparison of datasets in tests, make gnbsearchlight more inline with regular Searchlight

yarikoptic:
gnbsearchlight now would digest errorfx=None, and now pass cvfolds, targets to the result dataset


yarikoptic:
since no objections and it has been a while -- merging (or will forget about it again)

PR #329, opened 2015-06-12
ENH: Added passing fa to Projection mappers using Node.pass_attr

swaroopgj:
This mapper will allow adding feature attributes that will be propagated to datasets that are mapped using this mapper. It is useful to keep track of what the new features are after mapping.

### DIFF
--- a/mvpa2/base/node.py
+++ b/mvpa2/base/node.py
@@ -84,6 +84,8 @@ class Node(ClassWithCollections):
           'pvalues', while swapping the first and second axes. Simplified
           instructions can be given by leaving out consecutive tuple elements
           starting from the end.
+          Note: For Projection-derived mappers, this only passes attributes when
+          calling the object but not when `forward` is used to make output dataset.
         postproc : Node instance, optional
           Node to perform post-processing of results. This node is applied
           in `__call__()` to perform a final processing step on the to be
--- a/mvpa2/mappers/projection.py
+++ b/mvpa2/mappers/projection.py
@@ -14,7 +14,7 @@ import numpy as np
 
 from mvpa2.base.dochelpers import enhanced_doc_string
 from mvpa2.mappers.base import Mapper, accepts_dataset_as_samples
-
+from mvpa2.base.state import ConditionalAttribute
 
 if __debug__:
     from mvpa2.base import debug
@@ -43,6 +43,8 @@ class ProjectionMapper(Mapper):
     """
 
     _DEV__doc__ = """Think about renaming `demean`, may be `translation`?"""
+    
+    pass_fa = ConditionalAttribute(enabled=False, doc="Output feature attribute")
 
     def __init__(self, demean=True, **kwargs):
         """Initialize the ProjectionMapper
@@ -69,7 +71,8 @@ class ProjectionMapper(Mapper):
         """Offset (most often just mean) in the input space"""
         self._offset_out = None
         """Offset (most often just mean) in the output space"""
-
+        self.ca.pass_fa = None
+        """Feature attribute to be passed onto output dataset"""
     __doc__ = enhanced_doc_string('ProjectionMapper', locals(), Mapper)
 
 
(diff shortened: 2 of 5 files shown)


swaroopgj:
Any inputs on whether this feature (of adding fa in mapper) be part of current StaticProjectionMapper or a different mapper as is implemented currently?

hanke:
Sorry for the delay. It seems that the proposed functionality is very similar to what is done by the `pass_attr` argument that any `Node` supports. Do you need extra functionality that is not possible to implement with that facility?

swaroopgj:
Very good point. I think `pass_attr` might suffice. All I need to do is to enable a `ca` in `StaticProjectionMapper` that can be enabled at instantiation and assigned later. Is that true?

swaroopgj:
I experimented with adding `pass_fa` as `ca` in `ProjectionMapper`, but the attributes are only passed when the object is called, but not when using forward(). This makes sense, as the pass_attr is added in `Node._postcall()`. Is there a workaround to this, or do I need to manually call `Node._pass_attr` in `ProjectionMapper._forward_data()`?

hanke:
Hmm, not sure how to do this best.__call__() will execute __forward__() but
not vice versa. Any idea @yarikoptic ?

yarikoptic:
I do not really see a clean way out (yet at least). I guess as an ugly workaround we could overload _postcall in Mappers and move invocation of _pass_attr from there into some *forward

swaroopgj:
Will it be okay to call _pass_attr()  in forward() without changing anything else (for now)?

yarikoptic:
Stop -- why can't you just call the mapper, instead of an explicit .forward()?

swaroopgj:
Good point. Yes, we can just call the object instead. I use `forward` just because it seems intuitive in the code, and other users <i> might also do that </i>. Just to avoid confusion, we should make it explicit in the docs that `pass_attr` only works when calling the object and not with `forward` and people should use call and not forward, preferably. Does that sound good? If so, I can push my changes reflecting that.

yarikoptic:
On Fri, 26 Jun 2015, Swaroop Guntupalli wrote:

>    Good point. Yes, we can just call the object instead. I use forward just
>    because it seems intuitive in the code, and other users might also do that
>    . Just to avoid confusion, we should make it explicit in the docs that
>    pass_attr only works when calling the object and not with forward and
>    people should use call and not forward, preferably. Does that sound good?
>    If so, I can push my changes reflecting that.

sounds good to me!  please just verify that using .call() doesn't
introduce any side-effects ;)

mvpa2/base/node.py
```diff
@@ -84,6 +84,8 @@ def __init__(self, space=None, pass_attr=None, postproc=None, **kwargs):
           'pvalues', while swapping the first and second axes. Simplified
           instructions can be given by leaving out consecutive tuple elements
           starting from the end.
+          Note: For Projection-derived mappers, this only passes attributes when
```
yarikoptic: Well -- this would apply to any Mapper, not just Projection-derived, right?

mvpa2/mappers/projection.py
```diff
@@ -69,7 +71,8 @@ def __init__(self, demean=True, **kwargs):
         """Offset (most often just mean) in the input space"""
         self._offset_out = None
         """Offset (most often just mean) in the output space"""
-
+        self.ca.pass_fa = None
+        """Feature attribute to be passed onto output dataset"""
```
yarikoptic: oh hoh... hold on... now that I actually read your test which shows the use case.  This is an abuse of .ca which are intended for storage of computational artifacts, thus not to be assigned outside.

What you really want is just to assign that fa in your postproc which you could give to any node. e.g. you could say

```python
def myawkwardcustomfaassignment(ds):
  ds.fa['magicalone'] = ['whatever']*ds.nfeatures
  return ds
```
and then give  to your e.g. SVDMapper   postproc=myawkwardcustomfaassignment

or am I missing the point?

yarikoptic: ie, you don't need pass_attr at all... or am I missing the point?

swaroopgj:
You are correct. I can use postproc for my use case. Closing this PR.

PR #331, opened 2015-06-16
BF: Fixed the logic in checking that ref_ds param is in the correct range

swaroopgj:
Also added a unittest.

### DIFF
--- a/mvpa2/algorithms/hyperalignment.py
+++ b/mvpa2/algorithms/hyperalignment.py
@@ -198,7 +198,7 @@ class Hyperalignment(ClassWithCollections):
             ref_ds = np.argmax(nfeatures)
         else:
             ref_ds = params.ref_ds
-            if ref_ds < 0 and ref_ds >= ndatasets:
+            if ref_ds < 0 or ref_ds >= ndatasets:
                 raise ValueError, "Requested reference dataset %i is out of " \
                       "bounds. We have only %i datasets provided" \
                       % (ref_ds, ndatasets)
--- a/mvpa2/algorithms/hyperalignment.py
+++ b/mvpa2/algorithms/hyperalignment.py
@@ -124,7 +124,7 @@ class Hyperalignment(ClassWithCollections):
     level2_niter = Parameter(1, constraints=EnsureInt() & EnsureRange(min=0),
             doc="Number of 2nd-level iterations.")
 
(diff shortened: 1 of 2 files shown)


mvpa2/algorithms/hyperalignment.py
```diff
@@ -198,7 +198,7 @@ def train(self, datasets):
             ref_ds = np.argmax(nfeatures)
         else:
             ref_ds = params.ref_ds
-            if ref_ds < 0 and ref_ds >= ndatasets:
+            if ref_ds >= ndatasets:
```
nno: I like that it gives an informative error message for a too large value of ref_ds, but what about giving a non-integer or negative value?

swaroopgj: Good point. Those two check are done during declaration in Parameter() as EnsureInt() and EnsureRange(min=0)

nno: Ok, very good, thanks. How about adding a comment indicating that the Parameter class takes care of this? It may also be good to add two other cases to the test_hyper_ref_ds_range_checks function, using negative and non-integer ref_ds values.

swaroopgj: Current behavior accepts fractions that can be converted into non-negative ints (EnsureInt()). Would it be better to force the parameter to be of dtype int instead?

swaroopgj:
Any more comments on this fix?

nno:
- the current test code tests for an exception when ``__call__`` is used on an untrained Hyperalignment instance with ref_ds out of range. I haven't tried this but it seems an exception may be raised because the Hyperalignment instance is not trained, not because ref_ds is out of range. In any case, it would be good to have a test case where it checks an exception is thrown after training but with an out-of-range ref_ds value.
- could you add test cases verifying an exception is thrown negative and non-integer ref_ds

swaroopgj:
This exception should happen before training, which it does.
Negative and non-integer values are handled by Parameter class and it tries to cast them as Int with `EnsureInt()` if possible. I added a few lines to assert that.

yarikoptic:
better avoid magical casts (e.g. -0.5 -> 0), just give it ```EnsureInt(min=0)```

swaroopgj:
There is no such argument, but I go that functionality by switching constraints order. Also changed the test accordingly.

PR #333, opened 2015-06-23
ENH: Another update for the motionqc cmd

hanke:


mvpa2/datasets/sources/openfmri.py
```diff
@@ -315,6 +315,7 @@ def get_task_bold_attributes(self, task, fname, loadfx, exclude_subjs=None):
                 d = run_data[subj]
                 if len(d) == max_vol:
                     continue
+                # XXX WTF 6? determine shape of an attribute from other subjs
```
yarikoptic: agree ;-)

yarikoptic:
there is only so much we could do without tests for all this code :-/

PR #334, opened 2015-06-24
BF: minor doc fixups and PEP8

yarikoptic:


yarikoptic:
no objections -- merging ;)

PR #337, opened 2015-06-26
BF: allow specifying multiple additional regressors (1d or 2d) from .sa + making test faster

yarikoptic:


yarikoptic:
no objections -- merging

PR #339, opened 2015-06-28
Support storing into HDF5 ndarrays with unicodes

### LINKED ISSUE #335: Saving a niml.dset using h5save fails due to unrecognized dtypes(?)
I loaded a surface dataset from a .niml.dset file and niml.from_any() and it loads fine. But when I try to save the dataset using `h5save`, it spits out an error due to dtype of certain `sa` and  `a`. It looks like an issue with not handling unicode types, but I am not sure. Problematic dtypes look like '<U7' '<U4' when printing the ds. For now, I remove the attributes or convert them to strings before saving with h5save. @nno @yarikoptic

yarikoptic:
there is a bit of PEP8ing etc mixed in, sorry about that.

Closes #335

### DIFF
--- a/mvpa2/base/dochelpers.py
+++ b/mvpa2/base/dochelpers.py
@@ -560,6 +560,15 @@ def _str(obj, *args, **kwargs):
     return '<' + s + '>'
 
 
+def safe_str(obj):
+    """Return string of an object even if str() call fails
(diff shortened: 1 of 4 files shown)


nno:
Looks great, thanks!

PR #341, opened 2015-07-19
Improve fit2histogram + PEP8

hanke:


nno:
Thanks, looks good to me.

PR #343, opened 2015-07-20
support exporting tuple values from dataset .a to cosmo

nno:
minor NF to support CoSMoMVPA export of data in tuples in dataset .a. Also minor PEP8 cleanup thanks to PyCharm.

### DIFF
--- a/mvpa2/datasets/cosmo.py
+++ b/mvpa2/datasets/cosmo.py
@@ -144,6 +144,7 @@ Notes
 __docformat__ = 'restructuredtext'
 
 from mvpa2.base import externals
+
 externals.exists('scipy', raise_=True)
 
 from scipy.io import loadmat, savemat, matlab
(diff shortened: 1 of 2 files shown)


mvpa2/datasets/cosmo.py
```diff
@@ -323,15 +329,22 @@ def _attributes_dict2cosmo(ds):
...
+                        # transform vectors (shape=(P,)) to vectors in matrix
+                        # form (shape=(1,P))
+                        value = np.reshape(value, (-1, 1))
+
+                    if do_transpose:
+                        # feature attribute case, to match dimensionality
+                        # in second dimension
+                        value = value.T
+                except AttributeError:
+                    # not supported data element, skip
```
yarikoptic: shouldn't there be at least a warning???

nno: Not sure if a warning makes sense here. In the typical case of a dataset with a mapper, you would always get a warning. The current approach is just to silently ignore.

yarikoptic: well -- I see that it is to ignore here, but imho that is suboptimal since it hides away from user what is going on underneath -- some data is not preserved.  I will merge meanwhile but think about the warning ;-)

PR #345, opened 2015-07-20
RF(PEP8 etc): in externals code, to modernize it a bit as well

yarikoptic:
Let's wait for travis since locally I got failures -- yet to figure out if my code or not

### DIFF
--- a/mvpa2/base/config.py
+++ b/mvpa2/base/config.py
@@ -149,8 +149,10 @@ class ConfigManager(SafeConfigParser):
         class file2str(object):
             def __init__(self):
                 self.__s = ''
+
             def write(self, val):
                 self.__s += val
+
             def str(self):
                 return self.__s
 
@@ -180,11 +182,11 @@ class ConfigManager(SafeConfigParser):
 
         try:
             return SafeConfigParser.get(self, section, option, **kwargs)
-        except ValueError, e:
+        except ValueError as e:
             # provide somewhat descriptive error
-            raise ValueError, \
-                  "Failed to obtain value from configuration for %s.%s. " \
-                  "Original exception was: %s" % (section, option, e)
+            raise ValueError(
+                "Failed to obtain value from configuration for %s.%s. "
+                "Original exception was: %s" % (section, option, e))
 
 
     def getboolean(self, section, option, default=None):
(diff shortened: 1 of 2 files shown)


mvpa2/base/externals.py
```diff
@@ -438,7 +445,7 @@ def __check(name, a='__version__'):
         exec "v = %s.%s" % (name, a)
         # it might be lxml.etree, so take only first module
         versions[name.split('.')[0]] = SmartVersion(v)
-    except Exception, e:
+    except Exception as e:
```
nno: This syntax is not supported in Python 2.5.; the website mentions that Python 2.5 is supported.

mvpa2/base/config.py
```diff
@@ -180,11 +182,11 @@ def get(self, section, option, default=None, **kwargs):
 
         try:
             return SafeConfigParser.get(self, section, option, **kwargs)
-        except ValueError, e:
+        except ValueError as e:
```
nno: Not supported in Python 2.5

mvpa2/base/externals.py
```diff
@@ -208,8 +211,9 @@ def __check_shogun(bottom_version, custom_versions=[]):
     if (ver in custom_versions) or (ver >= bottom_version):
         return True
     else:
-        raise ImportError, 'Version %s is smaller than needed %s' % \
-              (ver, bottom_version)
+        raise ImportError(
+            'Version %s is smaller than needed %s'
+            % (ver, bottom_version))
```
nno: PEP8: "The preferred place to break around a binary operator is after the operator, not before it. "

mvpa2/base/externals.py
```diff
@@ -730,7 +728,7 @@ def check_all_dependencies(force=False, verbosity=1):
                 warning("%s is not available." % dep)
 
     if __debug__:
-        debug('EXT', 'The following optional externals are present: %s' \
-                     % [ k[5:] for k in cfg.options('externals')
-                            if k.startswith('have') \
-                            and cfg.getboolean('externals', k) == True ])
+        debug('EXT', 'The following optional externals are present: %s'
+                     % [k[5:] for k in cfg.options('externals')
```
nno: PEP8: "The preferred place to break around a binary operator is after the operator, not before it. "

hanke:
Let's not care about Python 2.5 anymore. Yes?

nno:
Fine with me not to support Python 2.5 anymore. Python 2.6 was released in 2008, almost 7 years ago, so I think it's fair to assume that almost (if not all) users use at least Python 2.6, and if not, it should be easy to upgrade.

mvpa2/base/externals.py
yarikoptic: oh -- pushed one more commit in docs to remove reference to 2.5 (in installation.rst) but may be there is more?  we have dumped 2.5 in 2012 iirc ;)

mvpa2/base/externals.py
yarikoptic: hm, interesting, forgot (happily) about that one.  IMHO this way is more descriptive, thus less bug prone -- you would immediately see what this line is for -- for string arguments, otherwise you need to jump to prev line to see that.  And PyCharm is not warning about this one, so I guess they also consider it good enough to ignore

yarikoptic:
ok, since no further objections -- merging

PR #346, opened 2015-07-20
set dataset attributes coming from NIML as vectors, not matrices (if possible)

nno:
Fixes an issue uncovered by #328 which added more stringent tests for dataset equality, causing test_surfing_afni.py to fail [1]. The unit test now also cleans up a temporary file after use.

[1] https://travis-ci.org/PyMVPA/PyMVPA/builds/71774997


nno:
@yarikoptic @hanke, we currently have a failing test on travis in the master branch. This PR should address this. Any objections to merging this?

hanke:
LGTM.

Issue #347, opened 2015-07-23
Support GIFTI I/O for datasets


nno:
Currently there is support for NIML but not GIFTI. It would be useful to add this using the nibabel.gifti.giftiio module.

nno:
Addressed through #348 and https://github.com/PyMVPA/PyMVPA/commit/05ebdda025401148425a7894b3a14ea73b932dfc

PR #348, opened 2015-07-26
add basic GIFTI support

nno:
- support GIFTI i/o for PyMVPA datasets
- adds new test function "assert_datasets_almost_equal"

### DIFF
--- a/mvpa2/datasets/gifti.py
+++ b/mvpa2/datasets/gifti.py
@@ -1,0 +1,190 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the PyMVPA package for the
+#   copyright and license terms.
+#
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""Support for magnetic resonance imaging (MRI) data IO.
+
+This module offers functions to import into PyMVPA surface-based GIFTI
+data using NiBabel_ (e.g. NIfTI, MINC, Analyze), and export PyMVPA
+surface-based datasets back into GIFTI.
+
+The current implementation supports data associated with nodes, and
+node indices for such data. There is no support for meta-data,
+non-identity affine transformations, or intent fields.
+
+This module supports node data, i.e. each node on the surface has N
+values associated with it (with N>=1). Typical examples include
+time series data and statistical maps. Other types of data, including
+label data or anatomical surfaces, are not supported.
+
+.. _NiBabel: http://nipy.sourceforge.net/nibabel
+"""
+
+
+from mvpa2.base import externals
+externals.exists('nibabel', raise_=True)
+
+from nibabel.gifti import gifti, giftiio
+
+from mvpa2.base.collections import FeatureAttributesCollection, \
+        SampleAttributesCollection
+from mvpa2.base.dataset import vstack
+from mvpa2.datasets.base import Dataset
+import numpy as np
+
+def _gifti_intent_niistring(intent_code):
+    return gifti.intent_codes.niistring[intent_code]
+
+def _gifti_intent_is_data(intent_string):
+    # exclude a set of intents
+    not_data_intent_strings=('NIFTI_INTENT_GENMATRIX',
+                             'NIFTI_INTENT_SYMMATRIX',
+                             'NIFTI_INTENT_DISPVECT',
+                             'NIFTI_INTENT_VECTOR',
+                             'NIFTI_INTENT_POINTSET',
+                             'NIFTI_INTENT_TRIANGLE',
+                             'NIFTI_INTENT_QUATERNION',
+                             'NIFTI_INTENT_DIMLESS',
+                             'NIFTI_INTENT_NODE_INDEX',
+                             'NIFTI_INTENT_SHAPE')
+
+    return intent_string not in not_data_intent_strings
+
+def _gifti_intent_is_node_indices(intent_string):
+    return intent_string == 'NIFTI_INTENT_NODE_INDEX'
+
+
+def _get_gifti_image(samples):
+    if isinstance(samples, basestring):
+        samples = giftiio.read(samples)
+
+    required_class=gifti.GiftiImage
+    if not isinstance(samples, required_class):
+        raise TypeError('Input of type %s must be a %s' %
+                                    (samples, required_class))
+
+    return samples
+
+def gifti_dataset(samples, targets=None, chunks=None):
+    """
+    Parameters
+    ----------
+    samples : str or GiftiImage
+      GIFTI surface-based data, specified either as a filename or an image.
+    targets : scalar or sequence
+      Label attribute for each volume in the timeseries.
+    chunks : scalar or sequence
+      Chunk attribute for each volume in the timeseries.
+    """
+    node_indices = None
+    data_vectors = []
+    intents = []
+
+    image = _get_gifti_image(samples)
+
+    for darray in image.darrays:
+        intent_string=_gifti_intent_niistring(darray.intent)
+
+        if _gifti_intent_is_data(intent_string):
+            data_vectors.append(darray.data)
+            intents.append(intent_string)
+
+        elif _gifti_intent_is_node_indices(intent_string):
+            node_indices=darray.data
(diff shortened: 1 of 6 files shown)


nno:
Currently fMRI datasets can also be written to GIFTI - which seems useless, because they cannot be visualised or analysed.
Would it make sense to let ``map2gifti`` check whether the dataset is fmri-like, and if so, throw an error?
@yarikoptic @hanke, any comments?

nno:
Addresses #347.

mvpa2/datasets/gifti.py
```diff
@@ -0,0 +1,190 @@
...
+        samples=ds.samples
+    else:
+        raise TypeError('first argument must be Dataset or numpy.ndarray')
+
+    [nsamples,nfeatures]=samples.shape
+
+    def _get_attribute_value(ds, attr_name, keys_):
+        try:
+            attr_collection=ds.__dict__.get(attr_name)
+        except:
```
yarikoptic: isn't it PEP8 or whatever to have at least "exception Exception"? but here it should be a `KeyError`?the point is that it might mask some other failure, not just that the key is not there

mvpa2/datasets/gifti.py
```diff
@@ -0,0 +1,190 @@
...
+
+    node_indices_labels = ('node_indices', 'center_ids', 'ids', 'roi_ids')
+    node_indices = _get_attribute_value(ds, 'fa', node_indices_labels)
+
+    if node_indices is not None:
+        darray = _build_array(node_indices,'NIFTI_INTENT_NODE_INDEX')
+        darrays.append(darray)
+
+
+    intents=_get_attribute_value(ds, 'sa', 'intents')
```
yarikoptic: PEP8: spaces around = ;-)

mvpa2/testing/tools.py
```diff
@@ -61,23 +62,28 @@ class SkipTest(Exception):
     assert_string_equal)
 
 
+
```
yarikoptic: even more spaces??? what for?? two empty lines is imho very sufficient -- think about fellas debugging in a 80x25 terminal remotely ;)

mvpa2/testing/tools.py
```diff
@@ -131,7 +162,7 @@ def assert_is_instance(obj, cls, msg=None):
     def assert_warnings(messages):
         with mock.patch("warnings.warn") as mock_warnings:
             yield
-            #import pydb; pydb.debugger()
+            # import pydb; pydb.debugger()
```
yarikoptic: next time if you see such a thing -- kill it! ;)

mvpa2/testing/tools.py
```diff
@@ -305,12 +350,15 @@ def newfunc(*arg, **kwargs):
                     # turn debug targets back on
                     debug.active = old_active
 
+
         newfunc = make_decorator(func)(newfunc)
         return newfunc
 
+
```
yarikoptic: please don't! ;)

mvpa2/tests/test_giftidataset.py
```diff
@@ -0,0 +1,185 @@
...
+
+        for format_ in ("ASCII", "Base64Binary", "GZipBase64Binary"):
+            # build GIFTI file from scratch
+            gifti_string = _build_gifti_string(format_, include_nodes)
+            with open(fn, 'w') as f:
+                f.write(gifti_string)
+
+            # reading GIFTI file
+            ds = gifti_dataset(fn)
+            assert_datasets_almost_equal(ds, expected_ds)
```
yarikoptic: how do you know that assert_datasets_almost_equal doesn't just not fail always? ;)  i.e. would be nice to have some unittest for it

mvpa2/tests/test_giftidataset.py
```diff
@@ -0,0 +1,185 @@
...
+@with_tempfile(suffix='.func.gii')
+def test_gifti_dataset(fn):
+    for include_nodes in (False, True):
+        expected_ds = _get_test_dataset(include_nodes)
+
+        expected_ds_sa = expected_ds.copy(deep=True)
+        expected_ds_sa.sa['chunks'] = [4, 3, 2, 1, 3, 2]
+        expected_ds_sa.sa['targets'] = ['t%d' % i for i in xrange(6)]
+
+        for format_ in ("ASCII", "Base64Binary", "GZipBase64Binary"):
```
yarikoptic: what about making it a parametric test , ie. moving all the looping outside.  Could be done either with 

```python
@sweepargs(include_nodes=(False,True))
@sweepargs(format_=("ASCII",  "Base64Binary", "GZipBase64Binary"))
@with_tempfile(suffix='.func.gii')
def test_gitfti_dataset(fn, format_, include_nodes):
   expected_ds = _get_test_dataset(include_nodes)
...
```
this will allow to fail/report only for specific combinations of options, and also avoid possible side-effects (e.g. if any function within inner loop modifies expected_ds or smth like that)

mvpa2/tests/test_giftidataset.py
```diff
@@ -0,0 +1,185 @@
...
+                         for sample in samples)
+
+    footer = """</GIFTI>"""
+
+    return header + "".join(body_elements) + footer
+
+
+
+@with_tempfile(suffix='.func.gii')
+def test_gifti_dataset(fn):
```
yarikoptic: could you please add testing of storing/loading of such datasets within mvpa2/tests/test_hdf5.py ?

yarikoptic:
On Mon, 27 Jul 2015, Nikolaas N. Oosterhof wrote:

>    Currently fMRI datasets can also be written to GIFTI - which seems
>    useless, because they cannot be visualised or analysed.
>    Would it make sense to let map2gifti check whether the dataset is
>    fmri-like, and if so, throw an error?
>    @yarikoptic @hanke, any comments?

may be  a warning would be sufficient?

mvpa2/tests/test_giftidataset.py
nno: How about adding some h5save/h5load to test_giftidataset? That way I don't have to duplicate code to generate the surface-based datasets.

mvpa2/testing/tools.py
nno: Hmm, this seems a limitation of PyCharm's autoformatting. There is one setting that applies to both top-level functions and classes at the same time. The code guidelines say to use 3 empty lines around classes but 2 around functions. Should we consider to have 2 lines also for classes?

nno:
I've tried to incorporate all suggestions made by @yarikoptic. In particular,
- unit tests for assert_datasets_equal and assert_datasets_almost_equal
- warning when attempting to convert volumetric dataset to GIFTI
- use of ``@sweepargs`` 
- test for h5save / h5load

mvpa2/tests/test_giftidataset.py
```diff
@@ -0,0 +1,204 @@
...
+        raise SkipTest
+
+    from mvpa2.base.hdf5 import h5save, h5load
+
+    ds = _get_test_dataset(include_nodes)
+
+    h5save(fn, ds)
+    ds2 = h5load(fn)
+
+    assert_datasets_almost_equal(ds, ds2)
```
yarikoptic: why not exactly equal? is there a loss in precision somewhere along the way?

mvpa2/tests/test_testing.py
```diff
@@ -100,4 +114,75 @@ def test_generate_testing_fmri_dataset(tempfile):
...
+
+    ds = Dataset(samples=samples, **args)
+
+
+    def negate_assert(f):
+        def raiser(*args, **kwargs):
+            assert_raises(AssertionError, f, *args, **kwargs)
+
+
+        return raiser
```
yarikoptic: screw pycharm - this is just insane... within 5 lines, 1 line of code and 4 of empty spaces -- what good does it do?

mvpa2/tests/test_testing.py
```diff
@@ -100,4 +114,75 @@ def test_generate_testing_fmri_dataset(tempfile):
...
+
+
+    def change_attribute(name, how_much):
+        # change a single attribute in samples, a, fa, or sa.
+        ds2 = ds.copy(deep=True)
+        attr = ds2.__dict__[name]
+        if name == 'samples':
+            value = attr
+        else:
+            for key in attr.keys():
```
yarikoptic: minor, feel free to ignore -- `.keys()` is not needed in such cases where attr is a dict or dict like. Makes it even "faster" since no copy of keys is created (in comparison to .iterkeys() in python2)

nno:
Thanks @yarikoptic for all the helpful comments, I tried to address them all.

PR #349, opened 2015-08-12
Enh fixups to improve daily_tests results (minor)

yarikoptic:


nno:
Thanks for the fixes, they look good to me.

PR #351, opened 2015-08-15
added more informative error msg

dinga92:
without this raise, it would crash on the next line with "TypeError: 'NoneType' object is not iterable" msg

### DIFF
--- a/mvpa2/generators/permutation.py
+++ b/mvpa2/generators/permutation.py
@@ -231,6 +231,9 @@ class AttributePermutator(Node):
 
     def _permute_chunks(self, limit_idx, in_pattrs, out_pattrs, chunks=None):
         # limit_idx is doing nothing
+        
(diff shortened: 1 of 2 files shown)


mvpa2/generators/permutation.py
```diff
@@ -231,6 +231,9 @@ def _permute_chunks_sanity_check(in_pattrs, chunks, uniques):
 
     def _permute_chunks(self, limit_idx, in_pattrs, out_pattrs, chunks=None):
         # limit_idx is doing nothing
+        
+        if chunks == None:
+            raise RuntimeError("Missing 'chunk_attr' for strategy='chunk'")
```
yarikoptic: comparison to none  better uses "is",  not ==
 and ValueError  would be  more  appropriate here  as well

 otherwise -  thanks for the PR  with the test!

PR #352, opened 2015-08-15
raise error instead of warning

dinga92:
Since it will not shuffle anything, there is no reason why anybody should use it despite the warning. Hence it should be an error instead


yarikoptic:
unit test?  :-)

PR #353, opened 2015-08-15
Permutator err msg

dinga92:


dinga92:
I just added the suggestions from #351, sorry for redundant PR

nno:
Thanks for the code. Before we merge:
- if #351 is redundant, can you please close it?
- the test fails because of a type mismatch of the error thrown: ValueError versus RuntimeError. Can you please update the test code?

yarikoptic:
coolio, so I have merged and fixed upthe test, and pushed.  So we are done with this one ;)

dinga92:
Great, thanks :)

PR #354, opened 2015-08-19
FIX: replaces deprecated fprob from scipy.stats with special.fdtrc

miykael:
According to [SciPy 0.16.0 Release Notes](https://github.com/scipy/scipy/blob/6aa03af69897582f290045b26b3a374b9fda29c0/doc/release/0.16.0-notes.rst#backwards-incompatible-changes), fprob was deprecated in version 0.16 and replaced with `special.fdtrc`, according to https://github.com/scipy/scipy/issues/2741.

I'm not sure if the term `fprob` should be replaced with `fdtrc` throughout PyMVPA.


yarikoptic:
Thanks!

Issue #342, opened 2015-07-20
Bug: Algorithm problem in SimpleSOMMapper


Soletmons:
I have been experimenting with the SimpleSOMMapper class, and came across what I think is a severe algorithm problem (please correct me if I'm wrong on this).

Currently, when training the SOM, the unit_delta to apply to each weight vector in the Kohonen layer is cumulatively calculated for each input training vector. This is then applied to the Kohonen layer only after every training vector has been considered. (see code snippet below)

This methodology has the unfortunate problem of creating very large numbers in the Kohonen layer. If a vector of weights is say 1 unit above all of a set of 1000 training vectors, the effect of one iteration will be to reduce the weights by 1000, where the desired effect is to reduce them by 1 (subject to learning rates/influence kernels). The next iteration, the weights will be increased by 1000*1000, and so on. Over many iterations, the Kohonen layer is driven to unreasonable values, breaking the algorithm. I believe this hasn't been spotted because the effect generally only becomes apparent for medium+ size sets of training vectors or iteration numbers, and the colours example used in the documentation is small.

The solution to this is to update the Kohonen layer each time a single training vector is considered (I'm happy to alter the code for this if you'd like).

It is also worth pointing out that the current implementation is different from some other versions of SOMs, in that often only one training vector is selected at random to apply on each iteration (here every training vector is used). Not necessarily a problem but important to say in the documentation.

Relevant code from SimpleSOMMapper:

        # for all iterations
        for it in xrange(1, self.niter + 1):
            # compute the neighborhood impact kernel for this iteration
            # has to be recomputed since kernel shrinks over time
            k = self._compute_influence_kernel(it, dqd)

            # for all training vectors
            for s in samples:
                # determine closest unit (as element coordinate)
                b = self._get_bmu(s)
                # train all units at once by unfolding the kernel (from the
                # single quadrant that is precomputed), cutting it to the
                # right shape and simply multiply it to the difference of target
                # and all unit weights....
                infl = np.vstack((
                        np.hstack((
                            # upper left
                            k[b[0]:0:-1, b[1]:0:-1],
                            # upper right
                            k[b[0]:0:-1, :self.kshape[1] - b[1]])),
                        np.hstack((
                            # lower left
                            k[:self.kshape[0] - b[0], b[1]:0:-1],
                            # lower right
                            k[:self.kshape[0] - b[0], :self.kshape[1] - b[1]]))
                               ))
[...]

PR #355, opened 2015-08-20
SOM bug fix and periodic boundary conditions

Soletmons:
There are two main changes:

-The Kohonen layer is updated with each sample, not just on each iteration. This prevents an issue where if the learning rate * the number of sample was > 1, the Kohonen layer was driven to extreme values.

-The Kohonen layer now has periodic boundaries. This is introduced through precomputing the influence array, then rolling it to the appropriate location for each sample.
(-) a new variable, self._dqdshape, is introduced to make computing the influence layer faster, as it now only needs to be calculated once per iteration and its shape does not change.

Both of these changes have a new test in test_som.py

Hopefully this helps! Apologies if I've done anything wrong, I'm quite new to github.

### DIFF
--- a/mvpa2/mappers/som.py
+++ b/mvpa2/mappers/som.py
@@ -88,6 +88,14 @@ class SimpleSOMMapper(Mapper):
         self._K = None
         self._dqd = None
         self._initialization_func = initialization_func
+        
+        # precompute necessary sizes for dqd (and later infl)
+        self._dqdshape = np.array(
+                             [self.kshape[0]/2,
+                              self.kshape[1]/2,
+                              np.ceil(self.kshape[0]/2.).astype('int'),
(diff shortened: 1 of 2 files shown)


hanke:
This looks good to me, thanks a lot. If there are no objections, I intend to merge this PR some time later today.

@Soletmons If you get a chance, please add yourself to the AUTHOR file. Thanks!

Soletmons:
Thanks! I've made a change to AUTHOR.

nno:
@Soletmons, thanks a lot for the work.
@hanke no objections, I will go ahead and merge.

This addresses #342. PR #356 was added to include David in AUTHOR, so I will merge that as well. 

@Soletmons, for future reference: you can update a current pull request by committing new changes and then push them again. In that way, a new PR is not needed very every update.

PR #357, opened 2015-08-25
ENH+BF: improve GIFTI functional data mapping support

nno:
This PR improves GIFTI support for functional data and fixes a bug:
- The GIFTI standard is not very clear about the presence of a coordinate matrix for data-arrays that are not point-sets, but FreeSurfer's mris_convert doesn't like their presence. This PR sets these coordinate matrices to None so that they are not stored.
- When storing functional data, is is cast to float32. Similarly, node indices are stored as int32
- map2gifti supports a "surface" argument to store anatomical information, so that FreeSurfer's mris_convert can deal with GIFTI files written by map2gifti

### DIFF
--- a/mvpa2/datasets/gifti.py
+++ b/mvpa2/datasets/gifti.py
@@ -176,8 +176,14 @@ def map2gifti(ds, filename=None, encoding='GIFTI_ENCODING_B64GZ'):
         return None
 
     def _build_array(data, intent, encoding=encoding):
-        return gifti.GiftiDataArray.from_array(data, intent,
-                                               encoding=encoding)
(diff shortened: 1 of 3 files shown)


nno:
This PR try to address a GIFTI issue raised on the mailinglist: http://lists.alioth.debian.org/pipermail/pkg-exppsy-pymvpa/2015q3/003231.html

yarikoptic:
Merging, but could you add a ChangeLog entry for this (and push directly).  Cheers!

nno:
Thanks. I just added this to the changelog file and pushed directly.

PR #361, opened 2015-09-06
Fancy print

dinga92:
Nice printing of dataset objects, so that samples and sa are in rows and features/fa are in columns. Just disclosing it here for a feedback, not intended for merge at the moment. For now it is using external library "tabulate" (pipable, MIT licence). 

usage:
ds.prt()
ds.prt(sa='targets')
ds.prt(sa=['targets', 'chunks'])

### DIFF
--- a/mvpa2/datasets/miscfx.py
+++ b/mvpa2/datasets/miscfx.py
@@ -425,7 +425,7 @@ def to_table(ds, sa=[], fa=[], height=100, width=10, n_digits=6):
 
     def create_attr_panel(ds_attr, keys, length):
         panel = []
-        panel.append(keys) # add header
+        panel.append(keys) #  add header
         for i in range(length):
             panel.append([ds_attr[key][i] for key in keys])
         return panel
@@ -461,6 +461,21 @@ def to_table(ds, sa=[], fa=[], height=100, width=10, n_digits=6):
 
 @datasetmethod
 def prt(ds, sa=[], fa=[], height=50, width=12, n_digits=5):
+    """Prints dataset object in table form
+
+    Parameters
+    ----------
+    sa : list or string, optional
+      Names of sample attributes to be displayed. Empty list will display all
+    fa : list or string, optional
+      Names of feature attributes to be displayed, Empty list will display all
+    height : int, optional
+      Number of rows printed
+    width : int, optional
+      Number of columns printed
+    ndigits : int, optional
+      Number of digits displayed in one cell
+    """
     height = min(ds.shape[0], height)
     width = min(ds.shape[1], width)
     print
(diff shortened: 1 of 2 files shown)


yarikoptic:
in   general i wholeheartedly support this feature! Could you include example in the pr  description  of  how it would look like?

dinga92:
of course I immediately found a bug and it doesn't work :)

mvpa2/tests/test_datasetfx.py
```diff
@@ -139,6 +142,80 @@ def test_sequence_stat(self):
...
+
+        # more fa + sa keys
+        table = ds.to_table()
+        correct_alignment(table, str.endswith)
+        correct_alignment(table.transpose(), str.startswith)
+
+        # Check if ds stays intact
+        table = ds.to_table(n_digits=1)
+        assert_array_equal(ds.samples, old_ds.samples)
+
```
yarikoptic: would be nice also to add a smoke test which would sweep through all test datasets and see if code doesn't blow. smth like
```python

@sweepargs(ds=datasets.itervalues())
def test_to_table_test_datasets(ds):
    dstxt = ds.to_table()
    for a in itertools.chain(ds.sa.keys(), ds.fa.keys(), ds.a.keys()):
         assert(a in dstxt)
```

mvpa2/datasets/miscfx.py
```diff
@@ -416,6 +418,71 @@ def cl_stats(axis, u, name1, name2):
...
+    width : int, optional
+      Number of columns printed
+    ndigits : int, optional
+      Number of digits displayed in one cell
+    """
+    height = min(ds.shape[0], height)
+    width = min(ds.shape[1], width)
+    print
+    print tabulate(to_table(ds, sa, fa, height, width, n_digits))
+    print "Ommitted %s rows and %s columns" %(ds.shape[0]-height, ds.shape[1]-width)
```
yarikoptic: instead of straight print better add kwarg "out=sys.stdout", and then use `out.write()` and append '\n' at the end for each line.  That would allow to use this function to dump output anywhere and also to test easily (which needs to be done)

mvpa2/datasets/miscfx.py
```diff
@@ -18,6 +18,8 @@
 
 import numpy as np
 
+from tabulate import tabulate
```
yarikoptic: this introduces additional external dependency (tabulate).  If sticking to it we would need to add conditioning on it being available, add its check within mvpa2.base.externals, etc

alternatively, we have similar one `table2string` but probably it is not as versatile and would require additional massaging of data before giving into it

yarikoptic:
@dinga92 interested to bring this baby to the finish line? ;)

hanke:
@dinga92 Ping! Let's get this done -- would be most appreciated!

PR #362, opened 2015-09-08
NF: Added alignment of datasets based on STATIS with tests

swaroopgj:


mvpa2/algorithms/statis.py
```diff
@@ -0,0 +1,142 @@
...
+        self.alpha = None
+
+    def __call__(self, dss):
+        """
+        Derive compromise matrix and aligned space.
+        :param dss: a list of datasets
+        :return:
+        """
+        if type(dss) == list:
+            print("Processing each of %d list items as tables"%(len(dss)))
```
nno: can this be a debug statement?

mvpa2/algorithms/statis.py
```diff
@@ -0,0 +1,142 @@
...
+            nss = [sd.nsamples for sd in dss]
+            if not nss.count(nss[0]) == len(nss):
+                raise ValueError, "All the datasets in the list should have matching" \
+                                "sample size. Input samples sizes are: %s" %nss
+        elif is_datasetlike(dss):
+            print("Processing each unique %s as table"%(self.tables_attr))
```
nno: debug statement?

mvpa2/algorithms/statis.py
```diff
@@ -0,0 +1,142 @@
...
+        self.outdim = min(nsamples-1, min(nfs))
+        # Deriving compromise matrix
+        cpms, C = get_Rv(dss)
+        e_t, ev_t = get_eig(C)
+        # XXX We can repeat the above two steps with permuted samples per column in each dataset
+        # to get a null distr of eigen values to evaluate how many are significant.
+        # First one should be, if second one is also, then there are two clusters.
+
+        # Factor scores for subjects/tables
+        self.G_t = np.dot(ev_t, np.diag(np.sqrt(e_t)))
```
nno: in DISTATIS, computing the compromise only makes sense if all pair-wise correlations between matrices from individuals are positively correlated. I don't know if that is also the case for STATIS... Is this something we may want to check, and throw an error if any pair is not positively correlated?

mvpa2/tests/test_statis.py
```diff
@@ -0,0 +1,104 @@
...
+from mvpa2.testing import sweepargs
+from mvpa2.testing.tools import assert_almost_equal
+class StatisTests(unittest.TestCase):
+    """
+    Adapted test from Hyperalignment testing as they both aim to
+    align datasets into common space
+    """
+    @sweepargs(hstacked=(True, False))
+    def test_basic_alignment(self, hstacked):
+        statis = Statis(tables_attr='subject')
```
nno: How about adding another test, using data from Abdi et al 2012, Table 1, and verify that the output matches the one in the paper?

mvpa2/algorithms/statis.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    Helper function to return sorted and sign flipped
+    Eigen decomposition.
+    :param C: Symmetric matrix
+    :return: eigen values, eigen vectors
+    """
+    evals, evecs = np.linalg.eigh(h)
+    # To deal with numerical errors
+    evals[evals<0] = 0
+    evecs = evecs[:, np.argsort(evals)[-1::-1]]
```
nno: how about storing the result from np.argsort, and use it to index evals and evecs? In that way there is only a single sort operation.

nno:
Thanks @swaroopgj, this looks very nice. I've added a few minor comments.

swaroopgj:
Added more tests and incorporated suggestions by @nno

mvpa2/base/__init__.py
```diff
@@ -347,6 +347,7 @@ def _set_max_count(self, value):
                    "needed for current refactoring carried out in tent/flexds")
 
     debug.register('BM', "Benchmark")
+    debug.register('STATIS', "Static algorithm")
```
yarikoptic: typo: Stati[s]

mvpa2/tests/test_statis.py
```diff
@@ -0,0 +1,125 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
```
yarikoptic: this file needs to be added now within `mvpa2/tests/__init__.py`

mvpa2/algorithms/statis.py
```diff
@@ -0,0 +1,218 @@
...
+class Statis(ClassWithCollections):
+    """Alignment of subjects or tables using STATIS
+
+    This algorithm takes a list of datasets with matching samples
+    or a single dataset with `fa.chunks_attr` separating subjects/tables.
+    It produces a list of `StaticProjectionMapper`s that project these feature spaces
+    of input datasets into STATIS compromise space.
+
+    Note that when provided a single dataset as input, returned mappers are in
+    the order fa.chunks_attr).
```
yarikoptic: widow ")"

mvpa2/algorithms/statis.py
```diff
@@ -0,0 +1,218 @@
...
+        # output space dimensionality
+        self.outdim = None
+        # Subject specific weights for compromise matrix
+        self.alpha = None
+
+    def __call__(self, dss):
+        """
+        Derive compromise matrix and aligned space.
+        :param dss: a list of datasets
+        :return:
```
yarikoptic: This is not the numpy convention we use... 
```
Parameters
------------

dss ...


Returns
----------
list of ...
```

mvpa2/algorithms/statis.py
```diff
@@ -0,0 +1,218 @@
...
+__all__ = ['Statis']
+
+def normalize_ds(ds):
+    """Normalize input dataset for Statis
+
+    1) zscore each column
+    2) make sum of squares to one (Frobenius norm = 1)
+    """
+    if ds.samples.dtype == int:
+        ds.samples = ds.samples.astype('float')
```
yarikoptic: this function is intended to return result dataset, not kosher to modify then original dataset inplace

check for datatype should allow for various int types, e.g. use
```
if np.issubdtype(ds.samples.dtype, np.integer):
```

yarikoptic: overall I would better just operate on a copy of the dataset...

```
ds = ds.copy()   # shallow copy to start with
if np.issubdtype(ds.samples.dtype, np.integer):
    ds.samples = ds.samples.astype('float')
else:
    ds.samples = ds.samples.copy()
zscore(ds)  # in-place zscore
```

mvpa2/algorithms/statis.py
```diff
@@ -0,0 +1,218 @@
...
+    zmapper = ZScoreMapper(chunks_attr=None)
+    zmapper.train(ds)
+    ds = zmapper.forward(ds)
+    ds.samples /= np.sqrt(((ds.samples**2).sum()))
```
yarikoptic: ah, so overall we are just doing computation of std and dividing by it in zscore for nothing....  why not to use zscore but demean manually inplace on the ds.samples and then do normalization?

mvpa2/algorithms/statis.py
```diff
@@ -0,0 +1,218 @@
...
+
+from mvpa2.base import warning
+if __debug__:
+    from mvpa2.base import debug
+
+# make only main utility visible outside
+
+__all__ = ['Statis']
+
+def normalize_ds(ds):
```
yarikoptic: since it doesn't do operation in place, why not name it "get_normalized_ds"?

mvpa2/algorithms/statis.py
yarikoptic: and ideally there should be unit-tests per each of the units -- i.e. of these helper functions

yarikoptic:
also list this as new feature in the ChangeLog please

swaroopgj:
Made changes as suggested by @yarikoptic

nno:
Thanks @swaroopgj.
However, the coverage has decreased substantially (7.6%), more than I would expect based on the amount of code added (with tests!). At the moment I don't have time to look into this, but it seems that is is not due to your changes; something else might be going on. Before we merge it would be good to figure out why the coverage has gone down by so much. @yarikoptic @hanke, any ideas on this?

yarikoptic:
Seems to be some fluke unrelated to these changes... Didn't look into details though to figure it out

yarikoptic:
Was about to merge but then -- is this really the Statis  or an alignment based on Statis (and thus better be StatisAlignment or smth like that)?  also add it to suite.py

swaroopgj:
It returns mappers for alignment. However, it does pretty much what is in the STATIS paper except plotting the results. The way I envisioned is the it returns mappers, but the object can be used to plot by storing required matrices. Clearly, more work needs to be added for that ;)

PR #363, opened 2015-09-17
DOC: adjust macports instructions. (Thanks Piotr T.)

yarikoptic:
Since I have no clue in macports -- I could be utterly wrong.  @nno?

### DIFF
--- a/doc/source/installation.rst
+++ b/doc/source/installation.rst
@@ -341,22 +341,21 @@ these changes to take effect.
 Once MacPorts is installed and your environment is properly configured, PyMVPA is
 installed using a single command::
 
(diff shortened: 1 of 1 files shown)


nno:
I'm not using macports either. In my setup I just installed anaconda, which brings nibabel, scipy etc, and then have PyMVPA through git. Then I used brew to install other packages, if needed. Maybe we should add the option of using anaconda / brew to the installation instructions?

yarikoptic:
Then I guess it couldn't become worse ;)  @nno -- have you had any experience packaging for (ana)conda?

nno:
@yarikoptic: I have no experience at all packaging for (ana)conda.

PR #365, opened 2015-09-29
Bug/cached query engine

bpinsard:
cached query engine misses the ids when trained


nno:
Thanks for the PR. To make sure that this would be fixed once and forever, could you please add a unit test so that:
- with the original code, the test fails due to missing query ids when using the GNBSearchlight
- with the new code, it passes. 
The test should probably also check the presence of the ids after training.
Cheers!

hanke:
@yarikoptic Objection to merge, or was this simply forgotten? LGTM.

nno:
Looks good to me, thanks.

Issue #368, opened 2015-10-03
travis coverage test is stalling


nno:
Recent merge of PR #367 seems to result in a consistent stall of travis for the coverage test (e.g. https://travis-ci.org/PyMVPA/PyMVPA/jobs/83428855), where no output is generated by nosetests for more than 10 minutes. It's unclear to my why it is stalling.
@yarikoptic @hanke, do you have any ideas about this?

nno:
It seems it's stalling due to externals check for ``me gt 2.4``: https://travis-ci.org/nno/PyMVPA/jobs/83560740

nno:
This report seems related: http://stackoverflow.com/questions/26547657/python-mdp-freeze-on-import-incompatibility-with-pp

nno:
Addressed by #369.

PR #369, opened 2015-10-05
ENH: use more recent mdp. 

nno:
This may fix an issue with travis hanging during coverage testing (#368).
@yarikoptic @hanke, any objections?


yarikoptic:
So they  addressed it already? Then we should cut a new mdp package for NeuroDebian

nno:
I'm not sure what exactly is the cause of travis stalling, and/or what change (if any) in mdp is responsible for this. I did not file a report to the mdb maintainers. 

After some debugging I just noted that using a recent mdp stops the stall on travis, whereas with the NeuroDebian version, recent runs on travis is associated with a stall.

yarikoptic:
OK, let's fix it this way then for his

nno:
Thanks!

PR #372, opened 2015-10-13
support VolumeMaskDict with different but compatible datatypes

nno:
When using surface-based voxel selection, sometimes auxiliary attributes in the VolumeMaskDict have different but compatible datatypes (such as float32 and float64). This PR allows for saving such VolumeMaskDict objects.
Thanks to @swaroopgj for reporting the issue.

This PR also includes some code cleanup following PEP8 guidelines.


nno:
@yarikoptic @hanke: @swaroopgj reported that the code works for him. Any objections to merging?

PR #373, opened 2015-10-13
rsa.PDTS multiple regression

rystoli:
Adds optional argument to rsa.PDistTargetSimilarity 'control_dsms', allowing RSA with multiple regression - looking at relation between neural and 'target DMs, controlling for other DMs (e.g., you may wish to control for multiple visual model DMs when looking at similarity structure in ventral temporal cortex). 

- 'control_dsms' is a list of squareformed DMs you wish to control for in a multiple regression. 

- Returns the r/rho coefficient for the target_dsm, controlling for any DMs in control_dsms. The r/rho coefficient returned is just the standard conversion of a OLS beta coefficient to an r coefficient for consistency in output ( beta * (sd(x) / sd(y)) ).

### DIFF
--- a/mvpa2/measures/rsa.py
+++ b/mvpa2/measures/rsa.py
@@ -22,6 +22,9 @@ if externals.exists('scipy', raise_=True):
     from scipy.spatial.distance import pdist, squareform
     from scipy.stats import rankdata, pearsonr
 
+if externals.exists('statsmodels', raise_=True):
+    from statsmodels.api import add_constant, OLS
+
 class PDist(Measure):
     """Compute dissimiliarity matrix for samples in a dataset
 
@@ -199,12 +202,15 @@ class PDistTargetSimilarity(Measure):
           If True, return only the correlation coefficient (rho), otherwise
           return rho and probability, p.""")
 
-    def __init__(self, target_dsm, **kwargs):
+    def __init__(self, target_dsm, control_dsms = None, **kwargs):
         """
         Parameters
         ----------
         target_dsm : array (length N*(N-1)/2)
           Target dissimilarity matrix
+        control_dsms : list of arrays (each length N*(N-1)/2)
+          Dissimilarity matrices to control for in multiple regression; flexible number allowed
+          *Optional. Returns r/rho coefficients for target_dsm, controlling for these dsms
 
         Returns
         -------
@@ -215,8 +221,10 @@ class PDistTargetSimilarity(Measure):
         # init base classes first
         Measure.__init__(self, **kwargs)
         self.target_dsm = target_dsm
+        self.control_dsms = control_dsms
         if self.params.comparison_metric == 'spearman':
             self.target_dsm = rankdata(target_dsm)
+            if control_dsms != None: self.control_dsms = [rankdata(dm) for dm in control_dsms]
 
     def _call(self,dataset):
         data = dataset.samples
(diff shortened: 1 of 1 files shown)


mvpa2/measures/rsa.py
```diff
@@ -199,12 +202,15 @@ class PDistTargetSimilarity(Measure):
           If True, return only the correlation coefficient (rho), otherwise
           return rho and probability, p.""")
 
-    def __init__(self, target_dsm, **kwargs):
+    def __init__(self, target_dsm, control_dsms = None, **kwargs):
```
yarikoptic: very minor nit-picking:  there should be no spaces around = within keyword definitions of 'def'.
Also all the optional parameters to this, as you can see above this line, specified as "Parameters".

yarikoptic: benefit is that it then automagically be listed by `__str__` and `__repr__` of this method

mvpa2/measures/rsa.py
```diff
@@ -215,8 +221,10 @@ def __init__(self, target_dsm, **kwargs):
         # init base classes first
         Measure.__init__(self, **kwargs)
         self.target_dsm = target_dsm
+        self.control_dsms = control_dsms
         if self.params.comparison_metric == 'spearman':
             self.target_dsm = rankdata(target_dsm)
+            if control_dsms != None: self.control_dsms = [rankdata(dm) for dm in control_dsms]
```
yarikoptic: correct way to check for None (or not) (here and below) is `control_dsms is not None`

mvpa2/measures/rsa.py
```diff
@@ -225,8 +233,19 @@ def _call(self,dataset):
...
-            return Dataset([[rho,p]], fa={'metrics': ['rho', 'p']})
+        if self.control_dsms == None:
+            rho, p = pearsonr(dsm,self.target_dsm)
+            if self.params.corrcoef_only:
+                return Dataset([rho], fa={'metrics': ['rho']})
+            else:
+                return Dataset([[rho,p]], fa={'metrics': ['rho', 'p']})
+        elif self.control_dsms != None:
+            X = np.column_stack([np.array([1. for i in self.target_dsm])]+[self.target_dsm]+self.control_dsms)
+            res = OLS(endog=dsm,exog=X).fit()
```
yarikoptic: to be kind to users, you could check if statsmodels was available (again) and if not -- spit out meaningful message, otherwise now it would just crash with "OLS not found"

yarikoptic: `[1. for i in self.target_dsm])` == `np.ones(len(self.target_dsm))`

mvpa2/measures/rsa.py
```diff
@@ -225,8 +233,19 @@ def _call(self,dataset):
...
+            rho = res.params[1]*(np.std(X[:,1])/np.std(dsm))
+            p = res.pvalues[1]
+            if self.params.corrcoef_only:
+                return Dataset([rho], fa={'metrics': ['rho']})
+            else: 
+                return Dataset([[rho,p]], fa={'metrics': ['rho','p']})
```
yarikoptic: This last block (`if self...`) is identical to if we have control_dsms or not, so move it out from the "if self.control_dsms is None`.  No need for elif here, else would suffice ;)

yarikoptic:
and now we would need to get some testing done -- look/extend `mvpa2/tests/test_rsa.py`.

hanke:
@rystoli Are you interested in finishing this PR up?

rystoli:
Hi all,

Sorry for the delay. 

We are trying to sort out whether we should set up something to take into account concerns of certain distance types linearly summing and if that should affect how neural distance is computed when using multiple regression (eg).

Also, I actually have to confess I'm at a loss for how exactly these tests are run that Yarik forwarded to me (he asked to to extend on these): mvpa2/tests/test_rsa.py. I am relatively amateur still with my coding chops and am not sure what / how to implement these. Any help would be appreciated!

Thanks!

rystoli:
example of the distance summation concern: http://biorxiv.org/content/biorxiv/early/2015/10/21/029603.full.pdf

PR #374, opened 2015-10-21
BF: correctly identify presence of IPython and behavior of SmartVersion if version=None was provided

yarikoptic:


nno:
Looks good, thanks.

PR #380, opened 2015-10-28
BF: import SurfaceQueryEngine into suite

yarikoptic:
@nno -- should be safe, right?


nno:
Very safe, I will merge. Thanks.

PR #382, opened 2015-11-03
RF: Remove deprecated nibabel get_* methods

effigies:
Nibabel deprecated the use of `img.get_header()` and `img.get_affine()` functions in favor of `img.header` and `img.affine` in [2.0.0](https://github.com/nipy/nibabel/tree/2.0.0). The `master` branch is now throwing a `DeprecationWarning` on `fmri_dataset` calls, so this patch replaces `.get_header()` (and `.get_affine()`, while I was at it).

In the process, I noticed that there are places that check for nibabel >= 1.2 to use `.shape` instead of `.get_shape()`. Would this be the preferred way to deal with the deprecations, or should these checks themselves be phased out?


yarikoptic:
Thank you @effigies !   Since nibabel 2.0 is generally available (and we have it throughout neurodebian for the releases where we also have recent mvpa2) we could just raise our versioned requirement  for it and just drop support for nibabel < 2, thus remove all those compatibility checks while at it.  Could you please remove those elderly checks, add a changelog entry for this change.

mvpa2/tests/test_niftidataset.py
```diff
@@ -113,34 +113,20 @@ def test_nifti_mapper(filename):
...
-    assert_equal(vol_shape, (128, 96, 24, 2))
+    assert_equal(vol.shape, (128, 96, 24, 2))
     ok_(isinstance(vol, getattr(nibabel, data.a.imgtype)))
 
     # test providing custom imgtypes
     vol = map2nifti(data, imgtype=nibabel.Nifti1Pair)
-    if externals.versions['nibabel'] >= '1.2':
-        vol_shape = vol.shape
-    else:
-        vol_shape = vol.get_shape()
```
effigies: This `vol_shape` seemed unused, so I didn't replace it.

effigies:
No rush, but anything you need from me on this PR?

yarikoptic:
thanks for the buzz -- didn't note further commits... thanks!

PR #383, opened 2015-11-13
enable pass_attr in simplestats searchlight when returning raw predictions

bpinsard:
If the partitions schemes doesn't output all samples in the same order for testing set, the passed attr either raise size mismatch or are incoherent.

### DIFF
--- a/mvpa2/measures/adhocsearchlightbase.py
+++ b/mvpa2/measures/adhocsearchlightbase.py
@@ -14,7 +14,7 @@ import numpy as np
 
 #from numpy import ones, zeros, sum, abs, isfinite, dot
 #from mvpa2.base import warning, externals
-from mvpa2.datasets.base import Dataset
+from mvpa2.datasets import Dataset, vstack
 from mvpa2.misc.errorfx import mean_mismatch_error
 from mvpa2.measures.searchlight import BaseSearchlight
 from mvpa2.base import externals, warning
@@ -425,7 +425,7 @@ class SimpleStatBaseSearchlight(BaseSearchlight):
             # Otherwise delay assembling the results
             results = []
 
-        all_targets, all_cvfolds = [], []
+        all_testing_sa, all_cvfolds = [], []
 
         # 4. Lets deduce all neighbors... might need to be RF into the
         #    parallel part later on
@@ -540,7 +540,7 @@ class SimpleStatBaseSearchlight(BaseSearchlight):
                 # and if no errorfx -- we just need to assign original
                 # labels to the predictions BUT keep in mind that it is a matrix
                 results.append(assign_ulabels(predictions))
-                all_targets += [ulabels[i] for i in targets]
+                all_testing_sa.append(split[1])
                 all_cvfolds += [isplit] * len(targets)
 
             pass  # end of the split loop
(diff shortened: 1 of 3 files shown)


mvpa2/measures/adhocsearchlightbase.py
```diff
@@ -555,12 +556,25 @@ def assign_ulabels(a):
...
         out.sa['cvfolds'] = all_cvfolds
         out.fa['center_ids'] = roi_ids
         return out
 
+    def _pass_attr(self, ds, result):
+        if self._all_testing_sa:
+            all_testing_sa = self._all_testing_sa
+            new_sas = {}
+            for attr in all_testing_sa[0].keys():
+                new_sas[attr] = np.concatenate([tsa[attr].value for tsa in all_testing_sa], 0)
```
yarikoptic: hm, so this one goes beyond of original pass_attr intention to just copy attributes from the original dataset, not to "track" them down all the way like this one does.  My concern is that it might be somewhat defeating the purpose of these adhoc searchlights to be fast -- in case of a dataset with heavy/lots of .sa's all the concatenation which would not get used (unless some pass_attr were specified) is somewhat wasteful :-/  At least condition it on having any 'self.pass_attr' defined and only then do this "copy everything" here and above... (otherwise -- only 'targets').  Actually ideally, at the top above (before looping) just call super's _pass_attr on an original ds, and mocked up result matching it to track which .sa would get copied, and then collect/concatenate only those to be passed along

mvpa2/measures/adhocsearchlightbase.py
```diff
@@ -555,12 +556,25 @@ def assign_ulabels(a):
...
+            result.sa['targets'] = new_sas['targets']
+            mock_attr_ds = Dataset(np.zeros((1,1)))
+            mock_attr_ds.sa = SampleAttributesCollection(new_sas)
+            mock_attr_ds.fa = ds.fa
+            mock_attr_ds.a  = ds.a
```
yarikoptic: cheater! ;)  you seems to exploit the fact/bug that we do not check dimensionality of .fa/.sa upon their assigning to the Dataset... interesting ;-)

mvpa2/tests/test_searchlight.py
```diff
@@ -762,6 +762,20 @@ def measure(ds):
...
+        # GNB searchlight with raw predictions
+        gnb_sl = GNBSearchlight(
+            GNB(),
+            generator=NFoldPartitioner(count=2),
+            qe=IndexQueryEngine(myspace=Sphere(2)),
+            errorfx=None,
+            pass_attr=ds1.sa.keys())
+        res = gnb_sl(ds1)
+        for k in ds1.sa.keys():
+            assert_in(k, res.sa.keys())
```
yarikoptic: heh, this test has no check whatsoever if concatenation worked out correctly.... you could have assigned some "sa.sample_id" with xrange and verify that obtained .sample_id is as it should be given ds.sa.chunks and the nfold partitioner with count=2

mvpa2/measures/adhocsearchlightbase.py
```diff
@@ -555,12 +556,25 @@ def assign_ulabels(a):
...
+            ds = mock_attr_ds
+        return super(SimpleStatBaseSearchlight, self)._pass_attr(ds, result)
```
yarikoptic: since iirc there should be no way this was invoked again without redoing training, probably reset that _all_training_sa to None whenever you are done with it here -- why to waste resources?

PR #384, opened 2015-11-13
Cv pass attr

bpinsard:
when cross-validating with partitions that doesn't use all dataset as testing or not in the same order than dataset samples, the passed attrs raise error for size mismatch or are either incoherent.


yarikoptic:
so seems to be pointing to the general flaw in the design that in many cases pass_attr should be either explicitly forbidden OR we would need to add some special ".sa._index_uid" and ".fa._index_uid" (where uid some random unique str) upon each invocation (if pass_attr defined) to verify if res has it identical to original, and if not -- kaboom! (not just silently ignore) since it points to the user's wrong assumption about what the beast was intended to do...

PR #386, opened 2015-11-16
ENH,WIP: add FactorialPartitioner to allow hierarchical cv

mvdoc:
With code from Yarik who created it in an afternoon of madness. Docstring still WIP, submitted to track it. Unittest is simply a check against the old chain of nfoldpartitioner and sifter.

### DIFF
--- a/mvpa2/generators/partition.py
+++ b/mvpa2/generators/partition.py
@@ -19,6 +19,8 @@ from mvpa2.base.node import Node
 from mvpa2.datasets.miscfx import coarsen_chunks
 import mvpa2.misc.support as support
 
+from itertools import product as iterprod
+
 if __debug__:
     from mvpa2.base import debug
 
@@ -489,6 +491,54 @@ class NFoldPartitioner(Partitioner):
             # right away
             return [(None, i) for ind, i in enumerate(combs)
                     if ind < self.count]
+
+
+class FactorialPartitioner(Partitioner):
+    """
+    Given another partitioner on a dataset with two hierarchically organized
+    attributes, generates the combinations of all possible folds of the
+    superordinate attribute with the balanced combinations of the subordinate
+    attribute. For example, splitting familiar and unfamiliar faces while
+    balancing identities in the training/test set.
+
+    ADD EXAMPLE HERE
+    """
+    def __init__(self, partitioner, **kwargs):
+        Partitioner.__init__(self, **kwargs)
+        # store the subordinate partitioner
+        self.partitioner = partitioner
+
+    def generate(self, ds):
+        class FakeDataset(object):
+            def __init__(self, ds):
+                self.sa = ds.sa
+
+            def copy(self, deep=False):
+                assert(not deep)  # dunno how
+                ds = FakeDataset(self)
+                ds.sa = ds.sa.copy(deep=False)
+                return ds
+
+            def __getitem__(self, item):
+                return FakeDataset(self[item])
+
+        #fakeds = FakeDataset(ds)
+        fakeds = ds[:, 0] # alternative
+
+        if self.selection_strategy != 'equidistant':
+            raise NotImplementedError("This strategy is not yet implemented")
+
+        attr_value = ds.sa[self.attr].value
+        uattr = ds.sa[self.attr].unique
+        uattr_masks = [attr_value == u for u in uattr]
+
+        for partitionings in iterprod(*[self.partitioner.generate(fakeds[uattr_mask]) for uattr_mask in uattr_masks]):
+            pds = ds.copy(deep=False)
+            target_partitioning = np.zeros(len(pds), dtype=int)
+            for uattr_mask, partitioning in zip(uattr_masks, partitionings):
+                target_partitioning[uattr_mask] = partitioning.sa[self.partitioner.space].value
+            pds.sa[self.space] = target_partitioning
+            yield pds
 
 
 class ExcludeTargetsCombinationsPartitioner(Node):
--- a/mvpa2/generators/partition.py
+++ b/mvpa2/generators/partition.py
@@ -16,6 +16,7 @@ from mvpa2.base.dochelpers import _repr_attrs
 from mvpa2.support.utils import deprecated
 
 from mvpa2.base.node import Node
+from mvpa2.base import warning
 from mvpa2.datasets.miscfx import coarsen_chunks
 import mvpa2.misc.support as support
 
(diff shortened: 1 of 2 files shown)


mvpa2/generators/partition.py
```diff
@@ -491,6 +493,54 @@ def _get_partition_specs(self, uniqueattrs):
                     if ind < self.count]
 
 
+class FactorialPartitioner(Partitioner):
+    """
```
yarikoptic: Provide one line summary description with empty line under,e.g.

```
"""Partitioner for two-level factorial designs

Given ...
"""
```

mvpa2/generators/partition.py
```diff
@@ -491,6 +493,54 @@ def _get_partition_specs(self, uniqueattrs):
...
+    Given another partitioner on a dataset with two hierarchically organized
+    attributes, generates the combinations of all possible folds of the
+    superordinate attribute with the balanced combinations of the subordinate
+    attribute. For example, splitting familiar and unfamiliar faces while
+    balancing identities in the training/test set.
+
+    ADD EXAMPLE HERE
```
yarikoptic: YES -- would be super useful.  Verify that it is "runnable" by running doctests

nosetests -s -v --with-doctest --doctest-tests mvpa2/generators/partition.py
```

File: mvpa2/generators/partition.py [hunk 3]
```diff
@@ -491,6 +493,54 @@ def _get_partition_specs(self, uniqueattrs):
...
+            def copy(self, deep=False):
+                assert(not deep)  # dunno how
+                ds = FakeDataset(self)
+                ds.sa = ds.sa.copy(deep=False)
+                return ds
+
+            def __getitem__(self, item):
+                return FakeDataset(self[item])
+
+        #fakeds = FakeDataset(ds)
```
yarikoptic: kill FakeDataset above since not used, and provide description under for why selecting a single feature

mvpa2/tests/test_generators.py
```diff
@@ -396,3 +397,46 @@ def is_sorted(x):
...
+    # now the new implementation
+    factpart = FactorialPartitioner(
+        NFoldPartitioner(attr='subord'),
+        attr='superord'
+    )
+
+    partitions_npart = [p.sa.partitions for p in npart.generate(ds)]
+    partitions_factpart = [p.sa.partitions for p in factpart.generate(ds)]
+
+    assert_array_equal(np.sort(partitions_npart), np.sort(partitions_factpart))
```
yarikoptic: we should test on few more cases, ideally corner cases.  E.g. What would happen if there is a single superordinate class?

would it blow or not, or issue a warning, if in one superord class we don't have all the subord categories? ;)

Also for completely dummy verification -- e.g. for 2x2 design verify that actual partitionings are as expected. (so we have a test which would not rely on correctness of NFold+Sifter construct operation)

mvpa2/generators/partition.py
```diff
@@ -491,6 +493,54 @@ def _get_partition_specs(self, uniqueattrs):
...
+    Given another partitioner on a dataset with two hierarchically organized
+    attributes, generates the combinations of all possible folds of the
+    superordinate attribute with the balanced combinations of the subordinate
+    attribute. For example, splitting familiar and unfamiliar faces while
+    balancing identities in the training/test set.
+
+    ADD EXAMPLE HERE
+    """
+    def __init__(self, partitioner, **kwargs):
+        Partitioner.__init__(self, **kwargs)
```
yarikoptic: to be safe, use `super` construct (even if other Partitioners use old fashion way like this, don't remember) -- we do have multiple inheritence around the hierarchy

mvpa2/tests/test_generators.py
mvdoc: Added test 1 and 3. For the second case, I tried and it doesn't work of course. My take is that it should blow when one superord class doesn't have all subord. What do you think?

mvpa2/generators/partition.py
```diff
@@ -491,6 +493,45 @@ def _get_partition_specs(self, uniqueattrs):
                     if ind < self.count]
 
 
+class FactorialPartitioner(Partitioner):
+    """Partitioner for two-level factorial designs
+
+    Given another partitioner on a dataset with two hierarchically organized
+    attributes, generates the combinations of all possible folds of the
+    superordinate attribute with the balanced combinations of the subordinate
+    attribute. For example, splitting familiar and unfamiliar faces while
```
yarikoptic: A bit more explicit, that splitting to cross validate across identities, because splitting faces is not clear... And that is under example. Use

```
Example
-----------
```

mvpa2/tests/test_generators.py
yarikoptic: I actually wonder... Probably it shouldn't blow if superords aren't balancedd in subords, since it might be a legit case and people might decide to use it with Balancer to mitigate

mvdoc: so just a warning?

yarikoptic: At most

mvdoc:
@yarikoptic let me know if this is ok

mvpa2/tests/test_generators.py
```diff
@@ -396,3 +397,79 @@ def is_sorted(x):
...
+
+    # now let's check it behaves correctly if we have only one superord class
+    nfold = NFoldPartitioner(attr='subord')
+    partitions_nfold = [p.sa.partitions for p in nfold.generate(ds_1super)]
+    partitions_factpart = [p.sa.partitions for p in factpart.generate(ds_1super)]
+    assert_array_equal(np.sort(partitions_nfold), np.sort(partitions_factpart))
+
+    # smoke test for unbalanced subord classes
+    partitions_factpart = [p.sa.partitions
+                           for p in factpart.generate(ds_unbalanced)]
```
yarikoptic: and what happens -- I guess warning is issued...  could you check if our assert_warnings works for this one? ;)
For that please do not use our ad-hoc mvpa2.base warning but `warnings.warn(msg, category=warnings.RuntimeWarning)` function, and here is an example from another place how to assert
```
from mvpa2.testing import assert_warnings
...
        with assert_warnings(
                [(DeprecationWarning,
                  "allowedtype option was deprecated in favor of constraints. "
                  "Adjust your code, provided value 'str' was ignored")]):
            DO_SOMETHIONG_WHICH_SPITS_WARNING
```

and what is the output? ;-) better to "hardcode" expectation here

Then I will become happy ;-)

mvdoc:
@yarikoptic better now? ;-)

PR #388, opened 2015-11-19
ENH: cmdline ttest supports hdf5

mvdoc:
Supports HDF5 for input datasets, mask, and output datasets


mvpa2/cmdline/cmd_ttest.py
```diff
@@ -58,14 +61,35 @@ def setup_parser(parser):
                         belong to.  E.g. if values are accuracies, it would be the
                         'greater', if errors -- the 'less'""")
 
+def check_filetype(fn):
+    if '.nii' in fn:
```
yarikoptic: Well... Nibabel and thus we support not only nifti... Moreover I don't see a point to have if s here and then in the originating function really at this point... What about just checking for common extensions of hdf and then if not matched giving to nibabel?

yarikoptic: Ah, II see the point for having a function now... So make it a guess_backend which would return hdf or nibabel ;-)

mvdoc:
ping @yarikoptic

mvpa2/cmdline/cmd_ttest.py
```diff
@@ -58,14 +61,38 @@ def setup_parser(parser):
                         belong to.  E.g. if values are accuracies, it would be the
                         'greater', if errors -- the 'less'""")
 
+def guess_backend(fn):
+    if fn.endswith('.gz'):
```
nno: How about an abstract class that has a static method guess_backend, and concrete Backend_* classes that provide methods for loading (e.g. nib.load) and getting data (e.g .get_data)? Possibly also a method indicating which extensions are supported.

yarikoptic: would it solve any specific problem or make code shorter? ;)

what I am saying, that In general -- it might be worth it, but not sure if really necessary for the scope of this PR.

nno: It might make it a bit shorter. Also probably a bit easier to understand.
There are now a few if-then-else statements that all deal with different backends. Why not use O-O to support the two backends?

mvpa2/cmdline/cmd_ttest.py
```diff
@@ -91,5 +118,13 @@ def run(args):
...
 
     verbose(1, "Saving to %s" % args.output)
-    nib.Nifti1Image(s, None, header=nis[0].header).to_filename(args.output)
+    filetype_out = guess_backend(args.output)
+    if filetype_out == 'nifti' and filetype == 'nifti':
+        nib.Nifti1Image(s, None, header=nis[0].header).to_filename(args.output)
+    else:
+        if filetype_out == 'nifti':
+            warning('Coercing output to hdf5 because input was hdf5')
+            args.output += '.hdf5'
```
yarikoptic: actually it might be quite cool if you could 
- check if there is a ds.a.mapper in one of the original datasets
- try to apply it map2nifti on result -- if succeeds -- no need to coerce, but can output nifti, if not -- better puke stating that requested file output type can't be obtained

mvdoc:
Still working on this.
@nno I understand your point, but to me it seems a bit overkill to implement an abstract class just for this cmdline option. If it were to be used in other cmds, then I agree it will be better to have a more robust implementation.

@yarikoptic suggestions on how to test this? I was thinking of mocking, or can we have dummy niftis/hdf5 datasets under tests?

mvdoc:
Travis seems to fail on `test_linear_sg` (which I didn't touch?) https://travis-ci.org/PyMVPA/PyMVPA/jobs/94981905#L2773

yarikoptic:
We have only few dummy niftis... I guess mocking wouldn't be that bad of an idea here - would make it fast ;-)

yarikoptic:
@mvdoc  as for testing -- you can easily generate at run time few tiny nifti and hdf5 files and test on those, that it behaves as desired and produces identical results either in hdf5 or nifti

yarikoptic:
Hi @mvdoc, could you please fix up for 
```
The command "eval   $COVERAGE $NOSETESTS -v mvpa2 $DOCTESTS -s" exited with 0.

0.18s$ make testsuite

I: Running full testsuite

E: 'from mvpa2.cmdline.cmd_ttest import' is missing from mvpa2.suite()

make: *** [testsuite] Error 1
```
causing travis to fail, merge master, and push again... if tested, I would like to merge it ;)

mvdoc:
```
FAIL: test_docstrings (mvpa2.tests.test_suite.SuiteTest)
  File "/home/travis/build/PyMVPA/PyMVPA/mvpa2/tests/test_suite.py", line 96, in test_docstrings
    self.fail('\n'.join(sfailures))
AssertionError: Some items have missing docstrings:
 functions: setup_parser, guess_backend, run
```

I recognize the beauty of this test. I'll fix it later

mvdoc:
@yarikoptic and this one?

mvpa2/cmdline/cmd_ttest.py
```diff
@@ -88,8 +122,31 @@ def run(args):
...
+            h5save(args.output, s_)
+    else:  # filetype_in hdf5
+        if filetype_out == 'nifti':
+            try:
+                map2nifti(dss[0], data=s).to_filename(args.output)
+            except (NameError, ValueError):
+                raise ValueError('Cannot output with requested file format')
+        else:  # filetype_out hdf5
+            s = Dataset(s, sa=dss[0].sa, fa=dss[0].fa, a=dss[0].a)
+            h5save(args.output, s)
```
yarikoptic: any chance to get this code tested?  ;) at least some smoke testing... so we don't blow up really badly? ;-)

mvpa2/tests/test_cmdline_ttest.py
```diff
@@ -0,0 +1,80 @@
...
+    assert_equal('hdf5', guess_backend('meh.h5'))
+    assert_equal('nifti', guess_backend('meh.tar'))
+
+# test that using a mask works
+datafn = pjoin(pymvpa_dataroot,
+               'haxby2001/sub001/BOLD/task001_run001/bold_25mm.nii.gz')
+maskfn = pjoin(pymvpa_dataroot,
+               'haxby2001/sub001/masks/25mm/brain.nii.gz')
+
+tmpdir = mkdtemp()
```
mvdoc: @yarikoptic this tmpdir doesn't get deleted ever. can I save a new hdf5 file under pymvpa_dataroot for later reuse?

yarikoptic: Ah... Missed this one... You need module level fixture, setup_module and teardown _module here where you would create and then remove anything... Just Google nose module fixture

mvpa2/tests/test_cmdline_ttest.py
```diff
@@ -0,0 +1,76 @@
...
+
+    args = FakeArg()
+    args.chance_level = 0.
+    args.alternative = alternative
+    args.stat = stat
+    args.mask = mask
+    args.data = [data for _ in range(3)]
+    args.output = outfn
+
+    run(args)
```
yarikoptic: Should be easy to check that it does save the file and it contains correct value, right? :-)

mvpa2/tests/test_cmdline_ttest.py
```diff
@@ -0,0 +1,76 @@
...
+            self.mask = ''
+            self.output = ''
+            self.isample = 0
+
+    args = FakeArg()
+    args.chance_level = 0.
+    args.alternative = alternative
+    args.stat = stat
+    args.mask = mask
+    args.data = [data for _ in range(3)]
```
yarikoptic: So easy to add some bogus known effect here... ;-)

yarikoptic: Ah, it is mocked and smoke anyways :-)

PR #392, opened 2016-01-04
Expose some functionality of group_clusterthr so could be used independently of the GroupClusterThreshold

yarikoptic:
We wanted to use parts of the implementation for similar cluster level thresholding but without bootstraps and other thrills.  While going through the logic, spotted one point to discuss (how cluster sizes get accumulated, around the point when [0] is returned) and provided few helper functions to work with those Counter's stats on cluster sizes collected, so they could be used easily to estimate p values for non-bootstrapped results.

### DIFF
--- a/mvpa2/algorithms/group_clusterthr.py
+++ b/mvpa2/algorithms/group_clusterthr.py
@@ -462,6 +462,12 @@ def get_thresholding_map(data, p=0.001):
 def _get_map_cluster_sizes(map_):
     labels, num = measurements.label(map_)
     area = measurements.sum(map_, labels, index=np.arange(1, num + 1))
+    # TODO: So here if a given map didn't have any super-thresholded features,
+    # we get 0 into our histogram.  BUT for the other maps, where at least 1 voxel
+    # passed the threshold we might get multiple clusters recorded within our
+    # distribution.  Which doesn't quite cut it for being called a FW cluster level.
+    # MAY BE it should count only the maximal cluster size (a single number)
+    # per given permutation (not all of them)
     if not len(area):
         return [0]
     else:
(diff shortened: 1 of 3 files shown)


mvpa2/algorithms/group_clusterthr.py
```diff
@@ -462,6 +462,12 @@ def get_thresholding_map(data, p=0.001):
 def _get_map_cluster_sizes(map_):
     labels, num = measurements.label(map_)
     area = measurements.sum(map_, labels, index=np.arange(1, num + 1))
+    # TODO: So here if a given map didn't have any super-thresholded features,
+    # we get 0 into our histogram.  BUT for the other maps, where at least 1 voxel
+    # passed the threshold we might get multiple clusters recorded within our
+    # distribution.  Which doesn't quite cut it for being called a FW cluster level.
+    # MAY BE it should count only the maximal cluster size (a single number)
+    # per given permutation (not all of them)
```
hanke: I agree. It should be the maximum cluster size, and zero if there was no cluster.

mvpa2/algorithms/group_clusterthr.py
```diff
@@ -505,6 +511,68 @@ def get_cluster_sizes(ds, cluster_counter=None):
...
+      dataset and null distribution
+    """
+    # TODO: dedicated unit-test for this function
+    """
+    Development note:
+     Functionality here somewhat dupliactes functionality in _transform_to_pvals
+     which does not operate on raw "Counters" and requires different input format.
+     Altogether with such data preparation _transform_to_pvals was slower than
+     this more naive implementation.
+    """
```
hanke: If it is slower, why keep it?

hanke:
OK to merge.

PR #393, opened 2016-01-05
WiP TST(BK): two use-case tests showing problems with adjusted StaticFeatureSelection

yarikoptic:
Two tests to demonstrate a problem with "adjusted" StaticFeatureSelection
- in first case it raises exception demanding training (IMHO shouldn't)
- in the 2nd case, with a ChainMapper it pretty much skips any reverse mapping

### DIFF
--- a/mvpa2/tests/test_mapper.py
+++ b/mvpa2/tests/test_mapper.py
@@ -499,8 +499,12 @@ def test_static_reverse_doesnt_work_after_feature_selection_tuneup_2():
     # would have gone all the way to 3d
     assert_equal(ds_orig_rev0.ndim, 3)
 
-    ds = ds_orig[:, [0, 2]]
-    ds_rev0 = ds.a.mapper.reverse1(ds.samples[0])
-    # blow test fails since chain mapper thrown exception upon _oshape mismatch
-    # but chain mapper silently proceeded forward doing nothing
-    assert_equal(ds_rev0.ndim, 3)
+    bool_mask = np.ones((ds_orig.nfeatures,), dtype=bool)
+    bool_mask[0] = False
+
+    for idx in (bool_mask, [0, 2]):
+        ds = ds_orig[:, idx]
+        ds_rev0 = ds.a.mapper.reverse1(ds.samples[0])
+        # blow test fails since chain mapper thrown exception upon _oshape mismatch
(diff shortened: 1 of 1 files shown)


mvpa2/tests/test_mapper.py
```diff
@@ -470,3 +470,41 @@ def test_identity_mapper(s):
...
+    ds = ds_orig.get_mapped(StaticFeatureSelection(np.arange(4)))
+    ds0_rev = ds.a.mapper.reverse1(ds.samples[0])  # should work
+    assert_equal(ds0_rev.shape, (1, ds_orig.nfeatures))
+
+    # direct feature selection
+    ds_ = ds[:, [0, 2]]
+    # should work but doesn't due to
+    # RuntimeError: Cannot reverse-map data since the original data shape is unknown. Either set `dshape` in the constructor, or call train().
+    ds0_rev_ = ds_.a.mapper.reverse1(ds_.samples[0])
+
```
hanke: Here is my version of the test -- which passes. The behavior (or lack thereof) of get_mapped() could be considered a bug.

````
def test_static_reverse_doesnt_work_after_feature_selection_tuneup_1():
    ds_orig = datasets['uni2small'].copy()  # doesn't matter which

    m = StaticFeatureSelection(np.arange(4))
    m.train(ds_orig)
    ds = ds_orig.get_mapped(m)
    ds0_rev = ds.a.mapper.reverse1(ds.samples[0])  # should work
    assert_equal(ds0_rev.shape, (ds_orig.nfeatures, ))

    # direct feature selection
    ds_ = ds[:, [0, 2]]
    # should work but doesn't due to
    # RuntimeError: Cannot reverse-map data since the original data shape is unknown. Either set `dshape` in the constructor, or call train().
    ds0_rev_ = ds_.a.mapper.reverse1(ds_.samples[0])
    assert_equal(ds0_rev.shape, (ds_orig.nfeatures, ))
````

Comment (yarikoptic) [inline mvpa2/tests/test_mapper.py]: aha -- tried to cheat me by my not so obscure use of "_"? ;)  last line should have been
```
assert_equal(ds0_rev_.shape, (ds_orig.nfeatures,))
```
and still not sure if magical autotraining should be done for the StaticFeatureSelection in this case

PR #395, opened 2016-01-06
ENH(WiP): perspective API improvement for GroupClusterThreshold

yarikoptic:
to support other metrics, QE, etc

wanted to see what parameters should come into play etc.  And it feels that may be there should be some class hierarchy since many parameters are specific for cluster forming (threshold, qe) and require the two-stage process (first estimate thr_map, and only then clusters for the metric).  In that two stage process also -- may be  we need change/parameter to remove a possible bias due to using the same bootstraps/data twice (I don't have a clear/strong feeling on that yet).

If we agree on basic idea/API -- I could prep the functionality and see how it goes from there (either requires classes etc)


mvpa2/algorithms/group_clusterthr.py
```diff
@@ -174,18 +174,40 @@ class GroupClusterThreshold(Learner):
...
 
     chunk_attr = Parameter(
-        'chunks',
+        'chunks', constraints=EnsureStr() | EnsureNone(),
         doc="""Name of the attribute indicating the individual chunks from
             which a single sample each is drawn for averaging into a bootstrap
-            sample.""")
+            sample.  If None, no bootstrapping will be performed, and each
+            sample will be used as a NULL result sample, thus there should be
+            a sufficient number of them""")
```
yarikoptic: this will allow for the same machine to be used if we have e.g. across subject classification results and e.g. 1000 permutations of those

Theoretically a plain NullDist machinery could be used BUT most probably it would blow, and it would not have all the clustering handling linked to it

yarikoptic: the point is that bootstrapping is just a first step needed only when we have insufficient number of NULL samples.  May be we could even manage to separate that out somehow into a generator which would furnish those samples to the actual logic

hanke: We just need to keep in mind that it should not become painful to generate the original samples adn the bootstrapped ones separately

mvpa2/algorithms/group_clusterthr.py
```diff
@@ -174,18 +174,40 @@ class GroupClusterThreshold(Learner):
...
+        doc="""What metric is measured per each value to be aggregated into
+            H0 distribution.  'max_cluster_size' collects a distribution of
+            maximal cluster size detected in a sample (or 0), 'cluster_sizes'
+            collects a distribution of all cluster sizes encountered while providing
+            a single 0 value per map if no clusters in that map were detected.
+            'cluster_sizes_non0' does not count 0s. 'max_value' collects a distribution
+            of maximal values.
+            Some metrics ('max_cluster_size', 'max_value') estimate family-wise
+            metric so no post-hoc correction is strictly necessary.""")
+
```
hanke: What does max_value do? This happens after thresholding, so if only max value is of interest, there is no need for thresholding at all.

mvpa2/algorithms/group_clusterthr.py
```diff
@@ -194,6 +216,18 @@ class GroupClusterThreshold(Learner):
...
+    qe = Parameter(
+        None,  #  constraints=  NOT SURE TODO
+        doc="""``QueryEngine`` (as used by searchlights) to determine feature's
+        immediate neighborhood.  If None provided, `ds.a.mapper` will be used to
+        reverse data into original possibly multi-dimensional shape.  If `ds`
+        does not have a mapper, 1D neighborhood along the features vector will
+        be used
+        """)
+    # TODO: relevant only for cluster-based analyses.
```
hanke: Would be lovely if it would work.

mvpa2/algorithms/group_clusterthr.py
```diff
@@ -294,6 +355,11 @@ def featuresegment_producer(ncols):
         # Step 2: threshold all NULL maps and build distribution of NULL cluster
         #         sizes
         #
+        # TODO: yoh: note that the same bcombos which were used to estimate thrmap
+        #       will now be used to estimate cluster sizes/metric.
+        #       Not sure if such double treatment of the same data doesn't provide
+        #       any bias.  We could  a) split chunks pool into two  b) create new
+        #       pool of bcombos for metric estimation
```
hanke: There is no double treatment. Usually one would generate the bootstrapped maps, threshold them, and determine cluster sizes. It is only done in two steps to avoid having to keep them and dramatically reduce the memory consumption.

mvpa2/algorithms/group_clusterthr.py
nno: @hanke a similar approach is used in CoSMoMVPA. But, one has to be careful that the query engine use the same space (e.g. voxels) and the same feature attributes. 
In CoSMoMVPA this is achieved by storing the .fa and relevant .a fields in the "query engine" (called neighborhood in CoSMo), so that when a neighborhood is used for clustering a dataset, it can be verified that these fields match.
See https://github.com/CoSMoMVPA/CoSMoMVPA/blob/master/mvpa/cosmo_cluster_neighborhood.m

nno:
Would this allow for TFCE and surface-based clustering?
As I wrote before, in my experience TFCE is much easier to implement if relevant cluster statistics are stored for each *feature* rather than each *cluster*. fixed-threshold clustering becomes then one special case.

yarikoptic:
@nno -- added `map_postproc` parameter to the beast, so we could feed it with any callable which transforms original or bootstrapped sample anyhow.  Implementation details yet to be figured out.  First I will implement clustering using QE (yet to comprehend what was your concern)

nno:
@yarikoptic, I did not have a particular concern - except that the current architecture does not seem easily extensible to support TFCE. But this architecture can be changed whenever there are plans to implement TFCE of course.

yarikoptic:
@PyMVPA/developers I have pushed my tentative RF to provide "clustering" using QEs.  The only additional somewhat demand is really that the dataset with permutations should carry a FlattenMapper somewhere with non-degenerate (i.e. non-None) space i.e. so we could identify which 'fa' to use for determining neighbors.  But that is only in the case if no explicit 'Labeler' (new construct implementing depth-first-search to identify clusters) was provided.  If provided - should be used.  Ideally we should get a unittest with data on surface.

I have disabled for now collection of 'spatial' statistics for clusters since that would require RFing of returned values anyways -- e.g. how should it look in case of surface dataset?  what if volumetric + temporal searchlight?  I thought that we should probably collect those using dataset's .fa and  query engine's arguments, but again not sure if those would always be consistent across all query engines.  IMHO such operation might better be performed explicitly after per each use case, e.g. `collect_spatial_clusters_stats(ds, fas=['voxel_indices'])`.  What do you think?

Also one test will break ATM since I was not sure on best way to address RFing of the internal `_get_map_cluster_sizes` which previously was taking a map and then calling .label() assuming pure Nd array to find neighbors in.  I could e.g. RF it so if no labels/num given it would flatten given data with our FlattenMapper and then proceed but imho it wouldn't be kosher.  I guess ideally I should provide it all those values so to test only that single invocation to count actual clusters ;)

What do you think?
Now I want to switch to enable other FW estimations, which wouldn't even care about "cluster" level anything

yarikoptic:
BTW @hanke , if you could check it out on some of your real results  -- that would be great.  It should provide identical results -- if not -- there is or was a bug ;)

hanke:
Doing the tests now....

First comment: The documentation needs to mention that in input dataset now needs to have a feature attribute indicating the spatial association of features. Before a mapper was sufficient -- which is now obsolete. This likely implies a change in existing analysis code, as the output of a searchlight, for example, does not have something like `voxel_indices`.

mvpa2/algorithms/group_clusterthr.py
```diff
@@ -462,22 +580,75 @@ def get_thresholding_map(data, p=0.001):
...
+    def __call__(self, map_, labels, num):
+        area = measurements.sum(map_, labels, index=np.arange(1, num + 1))
+        return self._metric(area)
+
+    def _cluster_sizes(self, area):
+        """Metric which for no clusters found returns [0]"""
+        if not len(area):
+            return [0]
+        else:
+            import pdb; pdb.set_trace()
```
hanke: Hu?

hanke:
@yarikoptic I will not keep a full analysis running to check for identical results. The estimated runtime of the cluster size calculation is 20 hours (up from ~10min; see #427 for performance estimates for this test analysis).

I will massively reduce the number of bootstrap samples and try again.

hanke:
@yarikoptic Apologies. It seems the new code behaves in non-obvious ways with parallelization and n_blocks>1. With n_proc=1 and n_blocks=1 it is wayyyyyy faster. Still playing...

mvpa2/measures/label.py
```diff
@@ -0,0 +1,107 @@
...
+                lmap[seed] = idx
+                while candidates:
+                    candidate = candidates.pop()
+                    # process its neighbors
+                    for neighbor in qe.query_byid(candidate):
+                        if (not d[neighbor]) or lmap[neighbor]:
+                            continue  # already labeled, so was considered etc
+                            # or simply was 0 to start with
+                        # immediately label
+                        lmap[neighbor] = idx
```
nno: Anything is a neighbor, even if .samples have different values? Why not support clustering so that two features are in the same cluster if they are connected and have the same value?

mvpa2/measures/label.py
```diff
@@ -0,0 +1,107 @@
...
+        super(Labeler, self).__init__(qe=qe, space=space, **kwargs)
+        self._untrain()
+
+    @property
+    def is_trained(self):
+        return self._nfeatures_trained is not None
+
+    def _train(self, ds):
+        self.params.qe.train(ds)
+        self._nfeatures_trained = ds.nfeatures
```
nno: Maybe it's safer to store the entire .fa? Then one could detect cases with the same number of features but still geometry.

hanke:
Ok, I am confused wrt performance. But anyway, here is a direct comparison of the original implementation and this PR without parallelization, and compartmentalization:

````
# ORIGINAL
# n_proc=1, n_blocks=1, n_bootstrap=5000
683.82s user 17.23s system 100% cpu 11:37.20 total

# THIS PR
# n_proc=1, n_blocks=1, n_bootstrap=5000
589.81s user 17.94s system 100% cpu 10:04.26 total
````

so performance looks all good in this case. However, the results differ. How much they differ is kinda hard to tell. The clusters in the new implementation are not sorted by size, so IDs cannot easily be matched, but it seems that the biggest clusters there is an approximate equivalent in size.

Looking at the results in terms of the location in the volume it seems that they are structurally the same. There are a few smaller clusters different between the two algorithms, but given the not-necessarily stellar number of bootstraps I consider this good enough.

I am in favor of replacing the current implementation with this more flexible one. But we should make sure to keep parallelization in mind. It could be that I screwed up somewhere, though....

**Comment by hanke** [comment]
Ah, and sorting the cluster IDs again will be much appreciated!

**Comment by hanke** [comment]
FTR: Before a merge the docstring needs a full pass to adjust to the new implementation.

File: mvpa2/algorithms/group_clusterthr.py [hunk 8]
```diff
@@ -462,22 +580,75 @@ def get_thresholding_map(data, p=0.001):
...
+            return []
+        else:
+            return area.astype(int)
+
+    def _max_cluster_size(self, area):
+        """Metric which returns maximum cluster size"""
+        if not len(area):
+            return [0]
+        else:
+            return area.astype(int)
```
hanke:
````
return area.astype(int).max()
````

?

File: mvpa2/algorithms/group_clusterthr.py [hunk 9]
```diff
@@ -462,22 +580,75 @@ def get_thresholding_map(data, p=0.001):
...
-    # we get 0 into our histogram.  BUT for the other maps, where at least 1 voxel
-    # passed the threshold we might get multiple clusters recorded within our
-    # distribution.  Which doesn't quite cut it for being called a FW cluster level.
-    # MAY BE it should count only the maximal cluster size (a single number)
-    # per given permutation (not all of them)
-    if not len(area):
-        return [0]
-    else:
-        return area.astype(int)
+class _ClustersMetric(object):
```
hanke: This should be `ClusterMetrics` instead.

mvpa2/algorithms/group_clusterthr.py
yarikoptic: D'oh -- thanks!  sure should have been a max somewhere ;)

yarikoptic:
@hanke  sorting by size should be inplace as of 9c7a3ff7063deb8b015bff9af40e928906878ba2 , note that differences still might happen if clusters have equal number of members.  Otherwise as of that commit I have refactored quite a few pieces primarily hopefully without change of functionality.  But now all 3 metrics for cluster level (size, sizenon0, maxsize) should work although still undertested I would say.

mvpa2/measures/label.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+class ClustersLabeler(Measure):
+    """Determines contiguous regions of features using a query engine.
+
+    "label" in the name chosen to mimic scipy.ndimage.measurements.label
+    """
+    qe = Parameter(
+        None,
+        doc="QueryEngine to determine neighbors of any given feature")
```
bpinsard: Will the cluster computation depend on the "radius" of the query engine? Or would it be possible to extract a neighborhood structure that takes only the closest neighboring samples?

yarikoptic: On Fri, 08 Apr 2016, bpinsard wrote:

>    In mvpa2/measures/label.py:

>  > +from ..misc.neighborhood import CachedQueryEngine
>  > +
>  > +from .base import Measure
>  > +
>  > +if __debug__:
>  > +    from ..base import debug
>  > +
>  > +
>  > +class ClustersLabeler(Measure):
>  > +    """Determines contiguous regions of features using a query engine.
>  > +
>  > +    "label" in the name chosen to mimic scipy.ndimage.measurements.label
>  > +    """
>  > +    qe = Parameter(
>  > +        None,
>  > +        doc="QueryEngine to determine neighbors of any given feature")

>    Will the cluster computation depend on the "radius" of the query engine?

if query engine takes radius, then yes ;)  but the point is to allow any
neighborhood querying by providing corresponding query engine.  E.g. in
the volume it could be the IndexQueryEngine which would take some
HollowSphere in one dimension (e.g. voxels coords) and then some other,
e.g. Sphere in another (e.g. time) thus allowing even to cluster
spatio-temporal beasts of arbitrary kinds ;)

>    Or would it be possible to extract a neighborhood structure that takes
>    only the closest neighboring samples?

define "samples" ;)  those query engines operate across "features" (in
pymvpa terminology), so if you need to operate across samples, you would
need to bring them into the same sample (e.g. boxcar'ing etc)

mvpa2/algorithms/group_clusterthr.py
```diff
@@ -236,205 +313,320 @@ def _train(self, ds):
...
-        # mapper default
-        mapper = IdentityMapper()
-        # overwrite if possible
-        if hasattr(ds, 'a') and 'mapper' in ds.a:
-            mapper = ds.a.mapper
-        # reverse-map input
-        othrd = _verified_reverse1(mapper, thrd)
-        # TODO: what is your purpose in life osamp? ;-)
-        osamp = _verified_reverse1(mapper, ds.samples[0])
+        thrd = ds_sample > thrmap
```
bpinsard: As I was previously computing error rate rather than accuracy, do you think of any case where negative tail would be of interest?
or both tails?

yarikoptic: so far I have just left logic on that identical to before, but indeed we could as well provide an option to instruct which tail to deal with...
interested to help with this PR? got distracted to other errands, never finished this PR, but it should be done

mvpa2/measures/label.py
bpinsard: Do you mean using the QE used to create the searchlight map?
It seems that there would be cases when multiple clusters are considered a single cluster just due to extended connectivity definition.  See the attached schema with thresholded map in blue and neighborhood in circles overlay.
slght
In that case it would certainly overestimate the null cluster size distribution by merging noisy patches in permutated maps, but also overestimate the cluster size in the true map by created superclusters?

Also when using surface QE, the neighborhood might vary for each subject for each location due to different distances between vertices of a standard surface spherically warped to the subject space. Which surface would you then use for group analysis, the template one (fsaveragex/icox)?

yarikoptic: 1. queryengine here is pretty much merely a generalization to define immediate neighborhood.  E.g. before `scipy.ndimage.measurements.label` internally was defining by default "immediate neighbors touching the face" and was restricted only to volume.  Here use of query engine relaxes it and allows to find an arbitrary immediate (or not if e.g. Sphere with big radius is used) neighborhood (in volume or on surface).  But in general/default you would just use `Sphere(1)` making it identical to how previous implementation behaved (using `scipy...label`)
2. interesting point on per-subject QEs!  Similar aspect was recently dealt with in searchlight hyperalignment (slhyper below) implementation (see e.g. https://github.com/PyMVPA/PyMVPA/blob/master/mvpa2/algorithms/searchlight_hyperalignment.py#L232 and where used in there).  But situation there was "easier" and somewhat clearer due to that slhyper obtaining a list of datasets, so they could easily carry varying .fa as per each subject and were not required to have corresponding features... Here we have inherent assumption that features already correspond across subjects, since we generally just computing a mean performance map, and clustering/qe is used later only to figure out clusters on that map.  So unless we want to aim for some extended functionality where QEs would also be used to project/mean performances within neighborhoods of each subject, there is probably no need for per-subject QE, but even then such meaning could simply be done prior, and for this computation clustering is done on e.g. a corresponding "template" surface

mvpa2/measures/label.py
yarikoptic: that might be a worthwhile feature, but do you see immediate (I love this word given that we last corresponded here over 3 years ago) use case?
The idea was -- to just pass a thresholded map (not the binarized mask) and do clustering.
In the long(er) run might indeed be nice to add an optional mode to take into account a value, so we could pass already some labeled maps, but I am yet to see the use-case

PR #398, opened 2016-01-10
Convenience wrapper around fmri_dataset()

hanke:
I am using code like this all over in my analyses. Before I work it in properly, I wanted to get a feel whether you would support merging this with the PyMVPA mainline.


mvpa2/datasets/mri.py
```diff
@@ -481,3 +481,77 @@ def strip_nibabel(ds):
...
+    ``fmri_dataset()``. Afterwards, additional sample attributes are assigned.
+    Lastly, the resulting (flattened and masked) dataset is subjected to
+    another pre-processing step that can be used to , for example, apply
+    temporal filtering.
+
+    Limitations with respect to ``fmri_dataset()`` exist insofar as this
+    function expects a single (4D) image to be specified, while calling
+    ``fmri_dataset()`` directly allows for specifying lists of individual
+    image (file names).
+
```
yarikoptic: Very minor - images
to specify

mvpa2/datasets/mri.py
```diff
@@ -481,3 +481,77 @@ def strip_nibabel(ds):
...
+    Returns
+    -------
+    Dataset
+    """
+    import nibabel as nb
+    # open the BOLD image
+    fmri_img = nb.load(fmri_fname)
+
+    if not preproc_img is None:
```
yarikoptic: Since I saw this pattern often. It is recommended (pep8?) To do 'var is not' and not 'not var is'

yarikoptic:
I don't see why this little function shouldn't have a little test ;-)

mvpa2/datasets/mri.py
yarikoptic: Are u sure this additional functionality shouldn't get absorbed into fmri_dataset?
Or alternatively, it could drop fmri_dataset specificity- provide that func as an argument and then could be used with any dataset constructor or factory

mvpa2/datasets/mri.py
hanke: My PEP8 checker is silent. Google brings up that some people feel that one is more natural than the other, but Python doesnt give a flying ... ;-)

````
>>> import dis
>>> def f(x):
...    return x is not None
...
>>> dis.dis(f)
  2           0 LOAD_FAST                0 (x)
              3 LOAD_CONST               0 (None)
              6 COMPARE_OP               9 (is not)
              9 RETURN_VALUE
>>> def g(x):
...   return not x is None
...
>>> dis.dis(g)
  2           0 LOAD_FAST                0 (x)
              3 LOAD_CONST               0 (None)
              6 COMPARE_OP               9 (is not)
              9 RETURN_VALUE
````

**Comment by hanke** [comment]
There is no test, because I am not sure whether there will be a function in the end. Hence the prolog to the PR.

File: mvpa2/datasets/mri.py (hunk 1 above)
Comment (hanke) [inline mvpa2/datasets/mri.py]: It would also need to drop nb.load... Not sure...

File: mvpa2/datasets/mri.py (hunk 2 above)
Comment (yarikoptic) [inline mvpa2/datasets/mri.py]: >    My PEP8 checker is silent. Google brings up that some people feel that one
>    is more natural than the other, but Python doesnt give a flying ... ;-)

https://www.quantifiedcode.com/knowledge-base/readability/Test%20object%20identity%20with%20%60is%60%20or%20%60is%20not%60/jYWiiO3O

-- 
Yaroslav O. Halchenko

File: mvpa2/datasets/mri.py (hunk 1 above)
Comment (hanke) [inline mvpa2/datasets/mri.py]: Maybe indeed. If I also pass nb.load-like functions as argument, I can use load_any_img and even support SPM-like file lists.... Maybe...

**Comment by hanke** [comment]
I pushed a more general implementation.

File: mvpa2/datasets/base.py [hunk 3]
```diff
@@ -561,3 +544,94 @@ def __getitem__(self, args):
...
+      that is understood by ``ds_converter`` (and any given ``preproc_raw``
+      callable).
+    ds_converter : callable
+      Callable that takes the output of ``raw_loader`` or ``preproc_raw``
+      as argument and returns a PyMVPA dataset.
+    preproc_raw : callable or None
+      If not None, this callable is used to perform initial preprocessing
+      after loading the data from its source. Must return data in a form
+      that is understood by ``ds_converter``.
+    preproc_ds : callable or None
```
yarikoptic: Looks good to me -- will merge now.  Just a feature request I guess.  Might be useful to support a "list of callables" for preproc_raw and preproc_ds (or may be any callable above really) so multiple operations could be chained as easily

Issue #381, opened 2015-11-01
numpy 1.10 compatibility


ignatenkobrain:
```python
======================================================================
ERROR: Test all classifiers for conformant behavior
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/PyMVPA-upstream-2.4.0/mvpa2/tests/test_clf.py", line 916, in test_generic_tests
    clf.train(traindata)
  File "/builddir/build/BUILD/PyMVPA-upstream-2.4.0/mvpa2/base/learner.py", line 132, in train
    result = self._train(ds)
  File "/builddir/build/BUILD/PyMVPA-upstream-2.4.0/mvpa2/clfs/meta.py", line 1371, in _train
    self.__mapper.train(dataset)
  File "/builddir/build/BUILD/PyMVPA-upstream-2.4.0/mvpa2/base/learner.py", line 132, in train
    result = self._train(ds)
  File "/builddir/build/BUILD/PyMVPA-upstream-2.4.0/mvpa2/featsel/base.py", line 320, in _train
    selected_ids = self._get_selected_ids(dataset)
  File "/builddir/build/BUILD/PyMVPA-upstream-2.4.0/mvpa2/featsel/base.py", line 296, in _get_selected_ids
    sensitivity = self.__sensitivity_analyzer(dataset)
  File "/builddir/build/BUILD/PyMVPA-upstream-2.4.0/mvpa2/base/learner.py", line 259, in __call__
    return super(Learner, self).__call__(ds)
  File "/builddir/build/BUILD/PyMVPA-upstream-2.4.0/mvpa2/base/node.py", line 121, in __call__
    result = self._call(ds)
  File "/builddir/build/BUILD/PyMVPA-upstream-2.4.0/mvpa2/measures/anova.py", line 88, in _call
    sostot /= bign
TypeError: ufunc 'divide' output (typecode 'd') could not be coerced to provided output parameter (typecode 'l') according to the casting rule ''same_kind''
```

ignatenkobrain:
The same was done in nipy projects, e.g. https://github.com/nipy/dipy/pull/731/

PR #400, opened 2016-01-11
BF: Achieve Numpy 1.10 compatibility (fixes gh-381)

hanke:
Casting rules changed.

### DIFF
--- a/mvpa2/measures/anova.py
+++ b/mvpa2/measures/anova.py
@@ -79,12 +79,12 @@ class OneWayAnova(FeaturewiseMeasure):
         ul = targets_sa.unique
 
         na = len(ul)
-        bign = float(dataset.nsamples)
         alldata = dataset.samples
 
         # total squares of sums
         sostot = np.sum(alldata, axis=0)
         sostot *= sostot
+        bign = sostot.dtype.type(dataset.nsamples)
         sostot /= bign
 
         # total sum of squares


hanke:
Tests are happy. Please merge @nno @yarikoptic

mvpa2/measures/anova.py
```diff
@@ -79,12 +79,12 @@ def _call(self, dataset):
         ul = targets_sa.unique
 
         na = len(ul)
-        bign = float(dataset.nsamples)
         alldata = dataset.samples
 
         # total squares of sums
         sostot = np.sum(alldata, axis=0)
         sostot *= sostot
+        bign = sostot.dtype.type(dataset.nsamples)
```
yarikoptic: this doesn't sound right -- if alldata.dtype is int, you will end up with int here, which would also point to the bug below that inplace division

```
$> python -c 'import numpy as np; a =np.arange(8).reshape((2, -1)); a = np.sum(a, axis=0); a*= a; print a.dtype.type; b=a.copy();  a /= 3; print a; print b/float(3)' 
<type 'numpy.int64'>
[ 5 12 21 33]
[  5.33333333  12.          21.33333333  33.33333333]
```
and I guess test isn't precise to care about those rounding effects

hanke: Ah, right. Didn't think about that. So we are left with a full copy on div...

yarikoptic: On Mon, 11 Jan 2016, Michael Hanke wrote:
>  >          alldata = dataset.samples

>  >          # total squares of sums
>  >          sostot = np.sum(alldata, axis=0)
>  >          sostot *= sostot
>  > +        bign = sostot.dtype.type(dataset.nsamples)

>    Ah, right. Didn't think about that. So we are left with a full copy on
>    div...

only if type is not float ;)

hanke:
Will merge this in -- have a bunch of fixes lined up and this is holding up the progress. Will deal with the fallout, if any.

yarikoptic:
Was it fixed for my concern of incorrect results if type was interested?

yarikoptic:
Bloody phone. int not interested

PR #401, opened 2016-01-11
WiP: NF: Utility command to make surface renderings

mih:
For those of us who are mentally incompatible with SUMA ;-)

Lots of docs missing, obviously. Question is: do we want such thing in PyMVPA at all?

### DIFF
--- a/bin/pymvpa2
+++ b/bin/pymvpa2
@@ -41,6 +41,7 @@ enabled_cmds = [
   'exec',
   'ofmotionqc',
   'ttest',
+  'rendersurf',
 ]
 
 # what version are we talking
--- a/mvpa2/cmdline/cmd_rendersurf.py
+++ b/mvpa2/cmdline/cmd_rendersurf.py
@@ -81,7 +81,7 @@ def setup_parser(parser):
         help="""""")
     parser.add_argument(
         '-r', '--overlay-range',
-        nargs=2, default=('min', 'max'),
+        nargs=2, default=('min', 'max'), type=float,
         help="""""")
     parser.add_argument(
         '-c', '--overlay-colormap', default='afmhot',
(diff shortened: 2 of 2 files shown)


yarikoptic:
we need one, but not only as cmdline utility, but also as a regular python function

advantage of @nno's one though is that it doesn't require any fancy 3d libraries, and theoretically could be improved to work on surfaces loaded straight from FS

hanke:
I assume you are referring to the flat surface plotter? I think these those are somewhat complementary. I rarely use flat maps and often prefer inflated (or partially inflated) surfaces, because I think they are easier to read.

Turning this whole thing into a big function is trivial.

Dependency issue is, and remains, a big bummer -- not much to argue...

yarikoptic:
yeap -- meant flat surface plotter. @nno -- any chance to get it fixed?

as for this one -- I don't mind any new functionality ;)

nno:
Thanks for the reminder @yarikoptic; I'll try to look into this in the next few days.

nno:
See #415 for improved surface plotting functionality.

hanke:
@yarikoptic @nno What about this one? Merge or reject?

yarikoptic:
does it work for you?  for travis? ;-)

hanke:
Works for me. Does travis come with Freesurfer? Should we install it prior
each test run? ;-)

yarikoptic:
for now we could adopt what pysurfer itself is doing:
```
    - wget --quiet http://faculty.washington.edu/larsoner/fsaverage_min.zip
    - mkdir subjects
    - cd subjects
    - unzip ../fsaverage_min.zip
    - cd ..
    - export SUBJECTS_DIR="${PWD}/subjects"
```

hanke:
That gives us minimal data, but if you run this command you'll see that it executes actual Freesurfer commands. It needs a full install. Maybe it could be worked around somehow with some effort, but for the 20 lines of plain pysurfer code, I'd rather close this PR unmerged than to spend a day figuring this out. TBH. No worries.

PR #402, opened 2016-01-11
ENH: allow limit for AttributePermutator to be a list

yarikoptic:
+ there is an additional commit with a little usecase test relying on that feature

### DIFF
--- a/mvpa2/algorithms/group_clusterthr.py
+++ b/mvpa2/algorithms/group_clusterthr.py
@@ -348,6 +348,7 @@ class GroupClusterThreshold(Learner):
             mapper = ds.a.mapper
         # reverse-map input
         othrd = _verified_reverse1(mapper, thrd)
+        # TODO: what is your purpose in life osamp? ;-)
         osamp = _verified_reverse1(mapper, ds.samples[0])
         # prep output dataset
         outds = ds.copy(deep=False)
--- a/mvpa2/generators/permutation.py
+++ b/mvpa2/generators/permutation.py
@@ -55,7 +55,7 @@ class AttributePermutator(Node):
           attribute name is given, its unique values will be used to define
           chunks of data that are permuted individually (i.e. no attributed
           values will be replaced across chunks). If a list given, then combination
-          of those attributes is used as a pair. Finally, if a dictionary is
+          of those attributes per each sample is used together. Finally, if a dictionary is
           provided, its keys define attribute names and its values (single value
           or sequence thereof) attribute value, where all key-value combinations
           across all given items define a "selection" of to-be-permuted samples
--- a/mvpa2/generators/permutation.py
+++ b/mvpa2/generators/permutation.py
@@ -50,11 +50,12 @@ class AttributePermutator(Node):
           all listed attributes.
         count : int
           Number of permutations to be yielded by .generate()
-        limit : None or str or dict
+        limit : None or str or list or dict
           If ``None`` all attribute values will be permuted. If an single
           attribute name is given, its unique values will be used to define
           chunks of data that are permuted individually (i.e. no attributed
-          values will be replaced across chunks). Finally, if a dictionary is
+          values will be replaced across chunks). If a list given, then combination
+          of those attributes is used as a pair. Finally, if a dictionary is
           provided, its keys define attribute names and its values (single value
           or sequence thereof) attribute value, where all key-value combinations
           across all given items define a "selection" of to-be-permuted samples
--- a/mvpa2/misc/support.py
+++ b/mvpa2/misc/support.py
@@ -18,6 +18,7 @@ import os
 from os.path import join as pathjoin
 import sys
 
+from itertools import product
 # for SmartVersion
 from distutils.version import Version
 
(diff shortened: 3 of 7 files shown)


yarikoptic:
I think test failure in 1 case was due to a labile test

yarikoptic:
any objections/recommendations @PyMVPA/developers ?

mvpa2/generators/permutation.py
```diff
@@ -50,11 +50,12 @@ def __init__(self, attr, count=1, limit=None, assure=False,
...
         count : int
           Number of permutations to be yielded by .generate()
-        limit : None or str or dict
+        limit : None or str or list or dict
           If ``None`` all attribute values will be permuted. If an single
           attribute name is given, its unique values will be used to define
           chunks of data that are permuted individually (i.e. no attributed
-          values will be replaced across chunks). Finally, if a dictionary is
+          values will be replaced across chunks). If a list given, then combination
+          of those attributes is used as a pair. Finally, if a dictionary is
```
nno: Somewhat unclear to me. Do you mean "If a list with n elements, then all possible combinations of those attributes are used as n-tuples"?

mvpa2/misc/support.py
```diff
@@ -740,12 +741,13 @@ def get_limit_filter(limit, collection):
...
-    limit : None or str or dict
-      If ``None`` all elements will be included in the filter. If an single
+    limit : None or str or list or dict
+      If ``None`` all elements will be included in the filter. If a single
       attribute name is given, its unique values will be used to define
-      chunks of data that are marked in the filter as unique integers. Finally,
-      if a dictionary is provided, its keys define attribute names and its
-      values (single value or sequence thereof) attribute value, where all
+      chunks of data that are marked in the filter as unique integers.
+      If a list given, then combination of those attributes is used as a pair.
```
nno: see above. Is it really "pair", or rather n-tuple?

mvpa2/tests/test_usecases.py
```diff
@@ -101,6 +89,27 @@ def test_sifter_superord_usecase():
     assert(np.mean(accs_regular) > .8)
     assert(np.mean(accs_super)   < .6)
 
+
+def _get_superord_dataset():
+    """A little helper to simulate a dataset with super/subord targets structure
+    """
+    # Let's simulate the beast -- 6 categories total groupped into 3
```
nno: typo

mvpa2/misc/support.py
yarikoptic: On Fri, 15 Jan 2016, Nikolaas N. Oosterhof wrote:

>    In mvpa2/misc/support.py:

>  >        attribute name is given, its unique values will be used to define
>  > -      chunks of data that are marked in the filter as unique integers. Finally,
>  > -      if a dictionary is provided, its keys define attribute names and its
>  > -      values (single value or sequence thereof) attribute value, where all
>  > +      chunks of data that are marked in the filter as unique integers.
>  > +      If a list given, then combination of those attributes is used as a pair.

>    see above. Is it really "pair", or rather n-tuple?

indeed... s/as a pair/together/

mvpa2/tests/test_usecases.py
yarikoptic: On Fri, 15 Jan 2016, Nikolaas N. Oosterhof wrote:

>    In mvpa2/tests/test_usecases.py:

>  > @@ -101,6 +89,27 @@ def test_sifter_superord_usecase():
>  >      assert(np.mean(accs_regular) > .8)
>  >      assert(np.mean(accs_super)   < .6)

>  > +
>  > +def _get_superord_dataset():
>  > +    """A little helper to simulate a dataset with super/subord targets structure
>  > +    """
>  > +    # Let's simulate the beast -- 6 categories total groupped into 3

>    typo

s/groupped/grouped/

mvpa2/generators/permutation.py
yarikoptic: On Fri, 15 Jan 2016, Nikolaas N. Oosterhof wrote:

>    In mvpa2/generators/permutation.py:

>  >            If ``None`` all attribute values will be permuted. If an single
>  >            attribute name is given, its unique values will be used to define
>  >            chunks of data that are permuted individually (i.e. no attributed
>  > -          values will be replaced across chunks). Finally, if a dictionary is
>  > +          values will be replaced across chunks). If a list given, then combination
>  > +          of those attributes is used as a pair. Finally, if a dictionary is

>    Somewhat unclear to me. Do you mean "If a list with n elements, then all
>    possible combinations of those attributes are used as n-tuples"?

added "per each sample" so it will sound as

If a list given, then combination
of those attributes per each sample is used together

mvpa2/algorithms/group_clusterthr.py
```diff
@@ -348,6 +348,7 @@ def _call(self, ds):
             mapper = ds.a.mapper
         # reverse-map input
         othrd = _verified_reverse1(mapper, thrd)
+        # TODO: what is your purpose in life osamp? ;-)
```
hanke: Sneaky comments are forbidden by PEP7653, sorry.

PR #405, opened 2016-01-16
Lots of tiny bits for general beautification

hanke:


hanke:
Even with coverage increase ;-)

Please merge.

nno:
Thanks @hanke

mvpa2/clfs/meta.py
```diff
@@ -1490,11 +1490,9 @@ def __init__(self, clf, centroids=None, distance_measure=None, **kwargs):
 
     def __repr__(self, prefixes=[]):
         if self.centroids is not None:
-            prefixes = prefixes + ['centroids=%r'
-                                   % self.centroids]
+            prefixes += ['centroids=%r' % self.centroids]
```
yarikoptic: This is not a mere beautification. It is changing into --inplace modification. Now try this on an object first which has centroids and then the one which didn't... I am afraid if would allow the bug. Copy prefixes first then at least

mvpa2/algorithms/benchmarks/hyperalignment.py
```diff
@@ -37,7 +37,7 @@ def wipe_out_offdiag(a, window_size, value=np.inf):
     value : optional
       Value to fill in with
     """
-    for r in range(len(a)):
+    for r, _ in enumerate(a):
```
yarikoptic: I would disagree that longer version is better here :-p

PR #406, opened 2016-01-16
BF: Replace all occurrences of mutable default values with a safe default

hanke:
Separate PR to get a clean test of just that.

### DIFF
--- a/mvpa2/atlases/base.py
+++ b/mvpa2/atlases/base.py
@@ -508,7 +508,9 @@ class LabelsLevel(Level):
 
 	XXX extend
     """
-    def __init__ (self, description, index=None, labels=[]):
+    def __init__ (self, description, index=None, labels=None):
+        if labels is None:
+            labels = []
         Level.__init__(self, description)
         self.__index = index
         self.__labels = labels
@@ -520,10 +522,12 @@ class LabelsLevel(Level):
 
     @staticmethod
     ##REF: Name was automagically refactored
-    def from_xml(Elevel, levelIndex=[0]):
+    def from_xml(Elevel, levelIndex=None):
         # XXX this is just for label type of level. For distance we need to ...
         # we need to assure the right indexing
 
+        if levelIndex is None:
+                levelIndex = [0]
         index = 0
         if 'index' in Elevel.attrib:
             index = int(Elevel.get("index"))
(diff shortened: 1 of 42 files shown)


nno:
Looks great, thanks.

mvpa2/atlases/base.py
```diff
@@ -520,10 +522,12 @@ def __repr__(self):
 
     @staticmethod
     ##REF: Name was automagically refactored
-    def from_xml(Elevel, levelIndex=[0]):
+    def from_xml(Elevel, levelIndex=None):
         # XXX this is just for label type of level. For distance we need to ...
         # we need to assure the right indexing
 
+        if levelIndex is None:
+                levelIndex = [0]
```
yarikoptic: Over indebted?

mvpa2/base/node.py
```diff
@@ -281,7 +281,9 @@ def set_postproc(self, node):
     def __str__(self, *args, **kwargs):
         return _str(self, *args, **kwargs)
 
-    def __repr__(self, prefixes=[]):
+    def __repr__(self, prefixes=None):
+        if prefixes is None:
+            prefixes = []
```
yarikoptic: It was perfectly fine before. More concise code and no chance for --inplace changes since +[]. Why to fix not broken?

yarikoptic:
@nno could we not rush a bit as much?

yarikoptic:
In this case I didn't like this one. Clearly many changes were due automatically and were not filtered out. In some places I would have preferred to keep it old way since it gives better idea on function signature, and had no side effect since that list was always added to another.  So why then to fix something not broken. Pep8 is not the law, it is a bible!

hanke:
This is not about PEP8. This is about the fragility that keep mutable argument "perfectly fine".

hanke:
In general those should be good and not hurt. Individual cases could be argued about, but then again: why make them special?

yarikoptic:
Once again in many places it was perfectly fine and code was written to account for this effect. Did you find a bug? (Besides the one introduced by beautification)?

yarikoptic:
Now many functions just look plain ugly

mvpa2/atlases/base.py
hanke: yes, sorry. These files have more formating issues though...

mvpa2/base/node.py
hanke: Speed accuracy trade-off. The fix is not that it was broken before. The fix is the function signature that looks more trustworthy now.

hanke:
I don't want to argue. Feel free to revert. However, personally I'd rather have it ugly like this than having to investigate each and every time whether the function body still reflect the implications imposed by a mutuable argument.

mvpa2/base/node.py
yarikoptic: I have no trust in function signatures. Moreover, the ones with all Nones in it defaults confuse me often

PR #407, opened 2016-01-16
RF: not X is None -> X is not None

hanke:
As it came repeatedly in the past as a style recommendation, this PR brings the current code base in line with that.


hanke:
So what about style PRs? ;-)

yarikoptic:
I haven't spotted any possible gotcha and tests passed, so let it be! ;)

PR #408, opened 2016-01-16
RM: Removal of Hamster()

hanke:
I don't think this class serves any purpose anymore. I haven seen it
being used in five years. I propose its removal.


hanke:
I guess this needs a joint vote with veto opportunity @yarikoptic @nno

nno:
With hdf5 I agree with @hanke that removal seems like a good idea.

yarikoptic:
+1 for RIP Hamster -- haven't used it forever

PR #409, opened 2016-01-16
BF: Remove a shadowed externals check to avoid confusion

hanke:
Could be better to do it the other way round though...


mvpa2/base/externals.py
```diff
@@ -573,7 +573,6 @@ def __check_liblapack_so():
           'reportlab': "__check('reportlab', 'Version')",
           'nose': "import nose as __",
           'pprocess': "__check('pprocess')",
-          'joblib': "__check('joblib')",
```
yarikoptic: travis failed... I guess it was the last one which had to go or what?

hanke:
I am not exactly sure why, but this issue is no longer present in master. Closing.

PR #412, opened 2016-01-16
ENH: Modernize SMLR test (run more of its code)

hanke:


yarikoptic:
in `__init__.py` then smmlr should get under nose list

yarikoptic:
or while you are into the "change everything" adjust tests/__init__.py to get rid of that separation and we could just demand nose.  Don't change all the tests though by dedenting them into functions from classes ;-)

hanke:
I am not in change anything mode anymore ;-) Added the test in __init__.py -- will merge once Travis gives green light.

PR #413, opened 2016-01-16
DOC: Who is using it.

hanke:
just a start. the list is long.


yarikoptic:
we are giving up on proper referencing?  
eventually I want to do https://github.com/duecredit/duecredit/issues/7 so we could just collate DOIs or BibTeX entries and get the list generated

hanke:
I am not giving up anything, but listing some in some way seems better than listing none at all. And so far we went for the latter for 2016 and virtually all of 2015.

yarikoptic:
@hanke -- I guess we could merge it now and then do it "properly" (all the references) later, right? ;)

PR #414, opened 2016-01-17
BF: Random attempt to fix the coverage setup

hanke:


hanke:
Worked now: +7.6%

Could not figure out the absolute root of the MDP installation, so I went for */src/mdp/mdp/* -- should be safe.

Also removed NotImplementedError lines from the report, as we do not (plan to) test them.

PR #416, opened 2016-01-18
BF+ENH: explicitly use float for "sum of" estimation in anova so we get proper float division

yarikoptic:
also OPT to use some inplace operations without temporary arrays being created

Since I haven't seen my concerns in https://github.com/PyMVPA/PyMVPA/pull/400 being addressed before/after merge and otherwise I would have forgotten.

Also added comparison against scipy implementation (which we should've just used at this point I guess)

### DIFF
--- a/mvpa2/measures/anova.py
+++ b/mvpa2/measures/anova.py
@@ -82,29 +82,34 @@ class OneWayAnova(FeaturewiseMeasure):
         alldata = dataset.samples
 
         # total squares of sums
-        sostot = np.sum(alldata, axis=0)
+        # determine data-dtype for derived "sum of" measurements, which we need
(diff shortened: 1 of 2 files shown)


hanke:
Err:
https://github.com/hanke/PyMVPA/commit/dea048349be768f2bcf54df0e1e8b77e683f53d5

I guess... I don't know...

There is another commit in that branch with more PEP8...

How do we want to proceeed?

yarikoptic:
didn't see it since it seems to be not a part of any PR (that numpy fix was merged before those changes were pushed)

yarikoptic:
@hanke I have updated my PR with merge of yours with my changes taking preference (theoretically should be faster due to inplace operations)

PR #418, opened 2016-01-22
ENH: make partitioner (aka generator) optional for CrossValidation and ad-hoc searchlights

yarikoptic:
generator=None  primarily to facilitate cross- classifications, so people could just specify
a Splitter on the values of the necessary attribute and be done

Also includes a bugfix for ad-hoc searchlights (such as gnbsearchlight) which disallowed to use custom partitioners which return more than 2 partitions.  While at it also made them work with a custom Splitter.

### DIFF
--- a/mvpa2/generators/partition.py
+++ b/mvpa2/generators/partition.py
@@ -130,7 +130,7 @@ class Partitioner(Node):
 
 
     def get_partitions_attr(self, ds, specs):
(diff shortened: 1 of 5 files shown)


yarikoptic:
@PyMVPA/developers any feedback?

hanke:
Like!

PR #419, opened 2016-01-25
BF: Sample attribute name convenience can fail for non-default settings

hanke:


mvpa2/mappers/glm/statsmodels_glm.py
```diff
@@ -58,5 +58,6 @@ def _fit_model(self, ds, X, reg_names):
                 add_constant=False,
                 model_gen=self.model_gen)
         res = mod(ds)
-        res.sa[self.get_space()] = reg_names
+        if self.result_expr == 'params':
```
yarikoptic: we have argument 'results' but storing as 'results_expr' attribute.  I wondered if we could unify one way or another.
Could there be a unittest which quickly demonstrated previous problem or too much hassle?

hanke: This is a complicated one, because it need to fit the Nipy one too. Also the current implementation needs to get the regressor names to statsmodels, otherwise it is very hard to use for contrasts.

If you don't mind, please merge without more tests. I'll go through the entire GLM-related code and refurbish things. Currently this is driven by what Nipy AND statsmodels can do -- quite limiting.

PR #421, opened 2016-01-27
NF: to/from_npz for Datasets

yarikoptic:
To ease data exchange with non-pymvpa-savvy users

One tricky part is that savez uses TEMPDIR so if dataset is large -- saving might fail if /tmp is not ample enough:  https://github.com/numpy/numpy/issues/5336


hanke:
This is useful!

PR #422, opened 2016-02-01
BF+ENH: set TR from scan_key for openfmri datasets' with TR=1 in nifti files

yarikoptic:
also added datalad external, if crawled openfmri ds 105 is present - smoke testing above
fix on it

also no dedicated unittest etc I would prefer to merge this one asap so ppl could start using OpenFMRIDataset construct with actual openfmri datasets.


nno:
Although slight decrease in coverage, it looks good otherwise to me. 
Since this seems somewhat urgent I will go ahead and merge, after I added a new issue #424 reminding to add a unit test.

PR #427, opened 2016-02-18
ENH: GroupClusterThreshold, some minor changes to get speedup 

mvdoc:
For the first step of training, compute number of blocks from n_proc. Then yield a numpy array and use np.percentile instead of argsort.

### DIFF
--- a/mvpa2/algorithms/group_clusterthr.py
+++ b/mvpa2/algorithms/group_clusterthr.py
@@ -26,6 +26,7 @@ from scipy.sparse import dok_matrix
 
 from mvpa2.mappers.base import IdentityMapper, _verified_reverse1
 from mvpa2.datasets import Dataset
+from mvpa2.base import externals
 from mvpa2.base.learner import Learner
 from mvpa2.base.param import Parameter
 from mvpa2.base.constraints import \
@@ -459,7 +460,13 @@ def get_thresholding_map(data, p=0.001):
     p_index = int(len(data) * p)
     if p_index < 1:
         raise ValueError("requested probability is too low for the given number of samples")
-    return np.percentile(data, q=(1-p)*100, interpolation='higher', axis=0)
+    if externals.versions['numpy'] >= '1.9.0':
+        return np.percentile(data, q=(1-p)*100, interpolation='higher', axis=0)
+    else:
+        # threshold indices are all in one row of the argsorted inputs
+        thridx = np.argsort(data, axis=0, kind='quicksort')[-p_index]
+        return data[thridx, np.arange(data.shape[1])]
+
 
 
 def _get_map_cluster_sizes(map_):
--- a/mvpa2/algorithms/group_clusterthr.py
+++ b/mvpa2/algorithms/group_clusterthr.py
@@ -263,13 +263,13 @@ class GroupClusterThreshold(Learner):
                 # one average map for every stored bcombo
                 # this also slices the input data into feature subsets
                 # for the compute blocks
-                yield [np.mean(
+                yield np.array([np.mean(
                        # get a view to a subset of the features
                        # -- should be somewhat efficient as feature axis is
                        # sliced
                        ds_samples[sidx, segstart:segstart + ncols],
                        axis=0)
-                       for sidx in bcombos]
+                       for sidx in bcombos])
         if self.params.n_proc == 1:
             # Serial execution
             thrmap = np.hstack(  # merge across compute blocks
@@ -319,7 +319,7 @@ class GroupClusterThreshold(Learner):
             # Parallel execution
             # same code as above, just restructured for joblib's Parallel
             for jobres in Parallel(n_jobs=self.params.n_proc,
-                                   #pre_dispatch=self.params.n_proc,
+                                   pre_dispatch=self.params.n_proc,
                                    verbose=verbose_level_parallel)(
                                        delayed(get_cluster_sizes)
                                   (Dataset(np.mean(ds_samples[sidx],
@@ -456,13 +456,10 @@ def get_thresholding_map(data, p=0.001):
       Value greater or equal than the returned threshold have a probability `p` or less.
     """
     # we need NumPy indexing logic, even if a dataset comes in
-    data = np.asanyarray(data)
     p_index = int(len(data) * p)
     if p_index < 1:
         raise ValueError("requested probability is too low for the given number of samples")
-    # threshold indices are all in one row of the argsorted inputs
-    thridx = np.argsort(data, axis=0, kind='quicksort')[-p_index]
-    return data[thridx, np.arange(data.shape[1])]
+    return np.percentile(data, q=(1-p)*100, interpolation='higher', axis=0)
 
 
 def _get_map_cluster_sizes(map_):
--- a/mvpa2/algorithms/group_clusterthr.py
+++ b/mvpa2/algorithms/group_clusterthr.py
@@ -287,7 +287,6 @@ class GroupClusterThreshold(Learner):
             # same code as above, just in parallel with joblib's Parallel
             thrmap = np.hstack(
                 Parallel(n_jobs=self.params.n_proc,
-                         #pre_dispatch=self.params.n_proc,
                          verbose=verbose_level_parallel)(
                              delayed(get_thresholding_map)
                         (d, self.params.feature_thresh_prob)
@@ -320,7 +319,6 @@ class GroupClusterThreshold(Learner):
             # Parallel execution
             # same code as above, just restructured for joblib's Parallel
             for jobres in Parallel(n_jobs=self.params.n_proc,
-                                   pre_dispatch=self.params.n_proc,
                                    verbose=verbose_level_parallel)(
                                        delayed(get_cluster_sizes)
                                   (Dataset(np.mean(ds_samples[sidx],
--- a/mvpa2/algorithms/group_clusterthr.py
+++ b/mvpa2/algorithms/group_clusterthr.py
@@ -245,7 +245,10 @@ class GroupClusterThreshold(Learner):
         #
         # Step 1: find the per-feature threshold that corresponds to some p
         # in the NULL
-        segwidth = ds.nfeatures / self.params.n_blocks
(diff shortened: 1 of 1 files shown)


mvpa2/algorithms/group_clusterthr.py
```diff
@@ -457,9 +461,13 @@ def get_thresholding_map(data, p=0.001):
     p_index = int(len(data) * p)
     if p_index < 1:
         raise ValueError("requested probability is too low for the given number of samples")
-    # threshold indices are all in one row of the argsorted inputs
-    thridx = np.argsort(data, axis=0, kind='quicksort')[-p_index]
-    return data[thridx, np.arange(data.shape[1])]
+    if externals.versions['numpy'] >= '1.9.0':
+        return np.percentile(data, q=(1-p)*100, interpolation='higher', axis=0)
```
nno: Is this really faster than the other solution in the 'else' block? By how much? If the difference is small, then adding more logic makes this harder to maintain.

mvpa2/algorithms/group_clusterthr.py
```diff
@@ -283,7 +287,7 @@ def featuresegment_producer(ncols):
             # same code as above, just in parallel with joblib's Parallel
             thrmap = np.hstack(
                 Parallel(n_jobs=self.params.n_proc,
-                         pre_dispatch=self.params.n_proc,
+                         #pre_dispatch=self.params.n_proc,
```
nno: Why is this commented out?

mvpa2/algorithms/group_clusterthr.py
```diff
@@ -260,13 +264,13 @@ def featuresegment_producer(ncols):
                 # one average map for every stored bcombo
                 # this also slices the input data into feature subsets
                 # for the compute blocks
-                yield [np.mean(
+                yield np.array([np.mean(
```
nno: Why the np.array constructor?

yarikoptic: On Tue, 23 Feb 2016, Nikolaas N. Oosterhof wrote:

>    In mvpa2/algorithms/group_clusterthr.py:

>  > @@ -260,13 +264,13 @@ def featuresegment_producer(ncols):
>  >                  # one average map for every stored bcombo
>  >                  # this also slices the input data into feature subsets
>  >                  # for the compute blocks
>  > -                yield [np.mean(
>  > +                yield np.array([np.mean(

>    Why the np.array constructor?

it seems that sending a single array through the pipe to joblib is
faster than sending a list of arrays which later get arrayed (in our
experimentation)

mvpa2/algorithms/group_clusterthr.py
mvdoc: because the default of pre_dispatch set to 3*n_proc should be a better option to avoid "starving" the cores. however it doesn't do much anything anyway in this case

mvpa2/algorithms/group_clusterthr.py
mvdoc: around 3 times faster

```python
In [7]: data = np.random.randn(100, 10000)

In [8]: %timeit np.argsort(data, axis=0, kind='quicksort')
10 loops, best of 3: 53.2 ms per loop

In [9]: %timeit np.percentile(data, axis=0, q=95, interpolation='higher')
100 loops, best of 3: 17.8 ms per loop
```

mvpa2/algorithms/group_clusterthr.py
nno: If it doesn't do much, can it be removed?

mvpa2/algorithms/group_clusterthr.py
nno: It may be 3 times faster, but (for data this size) in absolute terms it's only 40 ms faster. 
Is there reason to:
- expect data in real-world scenario's to be much bigger than this, leading to a significant larger delay using the original code; and
- assume that, for typical use case scenarios, a considerable amount of the total execution time of the main function is spent in these lines?
If not, I would vote for not including it.

I understand that this is a judgment call choosing between optimising and maintainability, and I appreciate the efforts for optimization; yet the speed benefit seems too small to justify maintaining and testing* two separate implementations.

[*] it would seem that testing two version of numpy would also involve quite a bit of extra effort.

mvdoc: Yeah, I see your point. Again, these are small changes that do not make the code extremely fast, but a little bit faster. The code should be refactored anyway because at this moment using parallelization makes the code slower rather than faster. It only gives a benefit for memory use. I was waiting for #395 to be merged before trying to refactor the parallelization loop.

hanke:
Looks good to me. I will re-run a full analysis with this today, and report any observations.

hanke:
Ok, here are a few datapoints to give some perspective on the starting point and the trajectory of the development. The following metrics are computed with a "real-world" analysis that I'd consider similar to a typical size: 11 subjects, ~240000 features, 150 raw permutation maps per subject as input. The implementation reference is PyMVPA 2.4.0 (precisely: abe024f59579698e60f4352c22b8729baeaa79c2)

````
# n_proc=1, n_blocks=24, n_bootstrap=5000
661.06s user 14.12s system 100% cpu 11:11.30 total
# n_proc=4, n_blocks=100, n_bootstrap=5000
940.18s user 128.54s system 198% cpu 8:59.55 total
# n_proc=12, n_blocks=50, n_bootstrap=5000
876.10s user 128.87s system 213% cpu 7:49.68 total
# n_proc=12, n_blocks=24, n_bootstrap=5000
python find_cluster_thresh_delay.py  861.47s user 128.70s system 216% cpu 7:37.01 total
````

it is not quite true that parallelization makes things slower, but I wouldn't want to call make faster either. Most importantly, it doesn't scale and is independent of the block size.

Next up: performance of this PR...

hanke:
Some performance metrics for this PR (rebased on the current master/master):

````
# n_proc=1, n_blocks=24, n_bootstrap=5000
682.42s user 17.75s system 100% cpu 11:36.53 total
# n_proc=4, n_blocks=100, n_bootstrap=5000
904.28s user 105.00s system 187% cpu 8:58.63 total
# n_proc=12, n_blocks=1, n_bootstrap=5000
852.13s user 117.26s system 186% cpu 8:40.99 total
# n_proc=12, n_blocks=24, n_bootstrap=5000
869.53s user 117.18s system 183% cpu 8:57.24 total
````

We cannot conclude much from these single-run estimates, but if it is moving at all, it doesn't seem to much in the right direction.

I should have said before that the reported runtimes correspond to a complete analysis script that also loads the data at the beginning (constant time of about 1min, machine has sufficient memory, I bet it didn't access the disks much...).

hanke:
Either I am doing something wrong or #395 also slows things done quite a bit (~20 hours runtime).

yarikoptic:
Slows down by 20 hours??? That would be yikes... There are some debug statement, could you please rerun with them enabled and reltime metric enabled so wet could see where it spent them to not be on a wild goose chase

hanke:
#395 is a different thing (and may be me). Let's keep the two PRs independent. And sorry for not having done that myself....

mvdoc:
@hanke thanks for running the performance metrics on a real analysis, it looks very...puzzling. just to clarify, in your second iteration using this PR, are you using `np.percentile` (i.e. does the machine have numpy >= 1.9.0?). 

Also, do you think it's worth trying to speed up this implementation, or better focus on #395?

hanke:
@mvdoc The second set of performance metrics was computed using this PR rebased against the current PyMVPA master -- so I guess it used np.percentile. Numpy was at version 1.8.2 (Debian stable). Machine specs: 2x 10-core 2.3 GHz Xeon E5-2650v3 96 GiB DDR4 RAM -- this much RAM was necessary to be able to compute n_blocks=1.

I think that we should move on with #395, make sure it doesn't slow things down, and then improve upon it.

The results are puzzling to me also. At first glance I cannot see why things should slow down. The only thing that concerns me is the additional array constructor, but if you say it become faster... Maybe your test arrays were small and this is a memory allocation/memcp issue? not sure....

yarikoptic:
On Sun, 06 Mar 2016, Michael Hanke wrote:

>    @mvdoc The second set of performance metrics was computed using this PR
>    rebased against the current PyMVPA master -- so I guess it used
>    np.percentile. Numpy was at version 1.8.2 (Debian stable). Machine specs:
>    2x 10-core 2.3 GHz Xeon E5-2650v3 96 GiB DDR4 RAM -- this much RAM was
>    necessary to be able to compute n_blocks=1.

>    I think that we should move on with #395, make sure it doesn't slow things
>    down, and then improve upon it.

>    The results are puzzling to me also. At first glance I cannot see why
>    things should slow down. The only thing that concerns me is the additional
>    array constructor, but if you say it become faster... Maybe your test
>    arrays were small and this is a memory allocation/memcp issue? not
>    sure....

I wonder if that it is indeed the case, especially if your niftis
originally were memmapped (i.e. come from .nii)?  then that
arraying/memcp would be detailed to each parallel subprocess making it
more efficient than doing serially in the main loop.

Primary advantage here was coming from doing it early on data which was
already in the memory (not memmapped), which seemed to be able to be
passed into subprocesses faster than a list of arrays.

So we might want to add there a logic to do early "arraying" only if
none of the underlying arrays is memmapped.  Then it would make this
heuristic efficient in both cases

-- 
Yaroslav O. Halchenko

hanke:
Data came from a bunch of hdf5s, and was compiled into a single dataset as part of the script. The chances of them still being memory-mapped are pretty low (lots of vstack-ing...)

mvdoc:
On Sun, Mar 6, 2016 at 6:32 AM, Michael Hanke <notifications@github.com>
wrote:

> Numpy was at version 1.8.2 (Debian stable).
>
@hanke So it must have used the old version with quicksort, because only
numpy >= 1.9.0 provides the option to choose how to solve a tie in a way
consistent with the quicksort solution. Still, it shouldn't be slower, but
approximately equal...

Anyway, as @nno was pointing out these (and arraying) are all minor tweaks
in the hope of speeding things up a little.

mvdoc:
I'm closing this since as @hanke pointed out better try #395 first.

PR #428, opened 2016-02-18
ENH: m1nn searchlight can now do correlation distance

yarikoptic:
As demanded by Sam, so for him to try/review/fix/improve

### DIFF
--- a/mvpa2/measures/nnsearchlight.py
+++ b/mvpa2/measures/nnsearchlight.py
@@ -16,7 +16,7 @@ import numpy as np
 from mvpa2.base.dochelpers import borrowkwargs, _repr_attrs
 from mvpa2.misc.neighborhood import IndexQueryEngine, Sphere
 
-from mvpa2.clfs.distance import squared_euclidean_distance
+from mvpa2.clfs.distance import squared_euclidean_distance, one_minus_correlation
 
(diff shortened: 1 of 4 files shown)


mvpa2/measures/nnsearchlight.py
```diff
@@ -44,7 +44,15 @@ def __init__(self, knn, generator, qe, **kwargs):
           Used to fetch space and dfx settings. TODO
         """
         # verify that desired features are supported
-        if knn.dfx != squared_euclidean_distance:
+        if knn.dfx == squared_euclidean_distance:
+            self._distance = 'euclidean'
+        elif knn.dfx == one_minus_correlation:
+            self._distance = 'correlation'
```
yarikoptic: and coveralls claims that this code is not tested, not sure how that could have happened

yarikoptic:
@snastase   so it worked without puking?  as we discussed it could be made more efficient through alternative way to compute it but not sure if I am to code it atm, so would like to merge

snastase:
@yarikoptic yes, works without puking and verified that it produces the same result as more generic (slow) kNN searchlight using one_minus_correlation dfx on a real fMRI dataset (up to a couple divide-by-zero errors)

yarikoptic:
good, merging then

Issue #430, opened 2016-02-25
Add support for CoSMoMVPA i/o of volumetric and surface-based fMRI data


nno:
As noted by @snastase and @mvdoc, currently exporting data from PyMVPA into CoSMoMVPA format requires manually changing .fa and adding .a.fdim in the output. I propose to extend map2cosmo with an argument 'data_type' which can be 'auto', 'fmri', 'surface', or None; the default would be 'auto'.
If 'fmri' or 'surface', then first it is verified that the dataset is indeed fMRI volumetric or surficial; if that is the case, the required .a.fdim and .fa attributes are updated. 
If 'auto', then it is detected whether the data is of type 'fmri' or 'surface', and then attributes are updated as above.

mvdoc:
:+1:  would love this ;-)

mvdoc:
Also @nno it seems there is a problem when one of the sample attributes contains strings. Here is a minimal example describing the problem:

```
In [6]: ds = Dataset([0, 1, 2], sa={'chunks': ['0+0', '1+1', '2+2']})

In [7]: ds
Out[7]: <Dataset: 3x1@int64, <sa: chunks>>

In [8]: cosmo.map2cosmo(ds)
Out[8]:
{'sa': array([[([['0+0'], ['1+1'], ['2+2']],)]],
       dtype=[('chunks', 'O')]), 'samples': array([[0],
        [1],
        [2]])}

In [9]: from scipy.io import savemat

In [11]: savemat('/tmp/test.mat', cosmo.map2cosmo(ds))
```

Then, in MATLAB

```
>> ds = load('/tmp/test.mat')

ds =

         sa: [1x1 struct]
    samples: [3x1 int64]

>> ds.sa

ans =

    chunks: [3x1x3 char]

>> ds.sa.chunks

ans(:,:,1) =

0
1
2

ans(:,:,2) =

+
+
+

ans(:,:,3) =

0
1
2
```

yarikoptic:
don't we have already .a.header providing that information for volumetric datasets already?  Don't remember details of surface ones atm. 
`and .fa attributes are updated` -- shouldn't happen in the provided dataset I guess -- side-effects better be avoided

nno:
[...]

PR #431, opened 2016-02-27
NF: support to export NIFTI to CoSMo format

nno:
WiP for #430 (@mvdoc @snastase). This first step allows exporting an fMRI dataset created in PyMVPA to CoSMoMVPA.

TODO for this PR: update CoSMoMVPA's ``cosmo_fmri_dataset`` so that the resulting struct can be read in CoSMoMVPA directly.

Roadmap: 
- implement i/o for the reverse way, i.e. allow an fmri dataset from CoSMoMVPA to be imported in PyMVPA, and be recognized as nifti-like frmi-dataset
- to do same for surface-based data (bi-directional)

### DIFF
--- a/mvpa2/datasets/cosmo.py
+++ b/mvpa2/datasets/cosmo.py
@@ -188,7 +188,6 @@ def _from_singleton(x, ndim=2):
                          (s, x, x.shape,))
     return x[(0,) * ndim]
 
-
 # dictionary indicating whether elements in .fa, .sa and .a must be
 # transposed for CoSMoMVPA data representation. As in CoSMoMVPA a dataset
 # with .samples PxQ the elements in .fa must be ?xQ, (only) .fa requires
@@ -327,7 +326,6 @@ def _attributes_dict2cosmo(ds):
             dtypes = []
             values = []
 
-
             for k in attr_collection:
                 value = attr_collection[k].value
 
@@ -383,8 +381,8 @@ def _mat_replace_functions_by_string(x):
     can write
     '''
 
-    unsupported_classes=(matlab.mio5_params.MatlabFunction,
-                         FlattenMapper)
+    unsupported_classes = (matlab.mio5_params.MatlabFunction,
+                           FlattenMapper)
 
     if isinstance(x, unsupported_classes):
         return np.asarray('%s' % x)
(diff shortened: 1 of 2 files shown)


nno:
The CoSMoMVPA-part has been merged into master. @yarikoptic @hanke, would you be available for a quick review?

mvpa2/datasets/cosmo.py
```diff
@@ -369,24 +372,28 @@ def _mat_replace_matlab_function_by_string(x):
...
     Notes
     -----
-    scipy can read but not write Matlab function hanndles; the use case of
+    scipy can read but not write Matlab function handles; the use case of
     this function is to replace such function handles by something that scipy
     can write
     '''
 
-    if isinstance(x, matlab.mio5_params.MatlabFunction):
+    unsupported_classes=(matlab.mio5_params.MatlabFunction,
```
yarikoptic: Pep8 spaces?
BTW I would recommend pycharm for python development ;-)

nno: I *am* using pycharm, and used its auto-format code option...

mvpa2/tests/test_cosmo.py
```diff
@@ -326,3 +359,34 @@ def test_cosmo_repr_and_str():
...
+        p_value = pymvpa_ds.a[k].value
+
+        if isinstance(p_value, tuple):
+            c_value = c_value.ravel()
+            p_value = np.asarray(p_value).ravel()
+
+        assert_array_almost_equal(c_value, p_value)
+
+    fa_dict=dict(_obj2tup(cosmomvpa_struct['fa']))
+    fa_keys=['voxel_indices']
```
yarikoptic: Pep 8 spaces

mvpa2/datasets/cosmo.py
yarikoptic: Ah cool
It must have warned about spaces here

yarikoptic:
If it works, then I guess everything is kosher. Eventually we should get some kind of integration test with full round trip to Cosmo and back

nno:
@yarikoptic It seems I did not run PyCharm auto-format on the latest code - it should be updated now, following PEP8 better.

yarikoptic:
Yeah since empty lines are gone! That is a fix on its own.
So do you want us to merge it now, out you want to get those to-do items into this pr before then?

nno:
@yarikoptic, thanks for the review. I guess we can merge now, and then do the other parts in a separate PR - that reduces the workload for the further pieces.

yarikoptic:
OK feel free to merge our I will do whether Travis passes again

nno:
@yarikoptic Thanks, there's no rush so let's just wait until Travis is happy

nno:
Thanks @yarikoptic.

PR #433, opened 2016-03-01
ENH: use rsync syntax compatible with what is present on OSX

nno:
on OSX, the syntax::

    rsync HOST::SRC1 HOST::SRC2 TRG 

is not supported; instead::

    rsync HOST::'SRC1 SRC2' TRG

must be used. This PR updates the ``fetch-data`` option in the Makefile that used the unsupported syntax.

(Unfortunately I did not find a way to split the source line into multiple lines)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,10 +19,10 @@ WWW_UPLOAD_URI_DEV=dev.pymvpa.org:/home/www/dev.pymvpa.org/pymvpa
 DATA_HOST=data.pymvpa.org
 DATA_DIR=datadb
 DATA_SUBDIRS='$(DATA_DIR)/tutorial_data \
-				 $(DATA_DIR)/mnist \
-				 $(DATA_DIR)/face_inversion_demo \
-				 $(DATA_DIR)/hyperalignment_tutorial_data \
-				 $(DATA_DIR)/haxby2001'
+                  $(DATA_DIR)/mnist \
+                  $(DATA_DIR)/face_inversion_demo \
+                  $(DATA_DIR)/hyperalignment_tutorial_data \
+                  $(DATA_DIR)/haxby2001'
 DATA_UPLOAD_URI=$(DATA_HOST):/home/www/data.pymvpa.org/www/datasets
 SWARMTOOL_DIR=tools/codeswarm
 SWARMTOOL_DIRFULL=$(CURDIR)/$(SWARMTOOL_DIR)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -707,12 +708,9 @@ bdist_mpkg: 3rd
 fetch-data:
 	echo "I: fetching data from datadb"
 	[ -e datadb ] || mkdir -p datadb
-	rsync $(RSYNC_OPTS) $(DATA_URI)/tutorial_data $(DATA_URI)/mnist \
-		$(DATA_URI)/face_inversion_demo \
-	      	$(DATA_URI)/hyperalignment_tutorial_data \
-                $(DATA_URI)/haxby2001 \
-		datadb/ 
-	@for ds in datadb/*; do \
+	rsync $(RSYNC_OPTS) $(DATA_HOST)::'$(DATA_DIR)/tutorial_data $(DATA_DIR)/mnist $(DATA_DIR)/face_inversion_demo $(DATA_DIR)/hyperalignment_tutorial_data $(DATA_DIR)/haxby2001' \
```
yarikoptic: But did it work on Linux's? Could it be multiplied as before may be?

nno: It seems to work fine on Linux (using hydra). What do you mean by "multiplied"?

yarikoptic: Multilined I meant... Damn phone

nno: ``> Damn phone``
Isn't that a phone that beats iphone? :-)

I tried multi-lined, but did not find a solution. When using::

    'a \
    b'

this is converted to 'a \b', and I did not find a way to get rid of the backslash. If you have any suggestion on how to do this I would really appreciate it.

yarikoptic: I don't think you would need \ or it didn't understand it then?

nno: I tried (but failed) to split the part between single quotes at line 711 into multiple lines:: 

    '$(DATA_DIR)/tutorial_data $(DATA_DIR)/mnist $(DATA_DIR)/face_inversion_demo $(DATA_DIR)/hyperalignment_tutorial_data $(DATA_DIR)/haxby2001'

Any ideas how to do that, or is it just not possible?

yarikoptic: Only in the phone ATM so can't try. Alternative is may be too place all those into a variable and there you could may be list them in multiple lines with \

yarikoptic:
ok -- this seems to work
```
$> make
echo rsync -a localhost::"/tmp/f1 /tmp/f2 /tmp/f3" ...
rsync -a localhost::/tmp/f1 /tmp/f2 /tmp/f3 ...

$> cat Makefile 
var=/tmp/f1\
 /tmp/f2\
 /tmp/f3

all:
	echo rsync -a localhost::"$(var)" ...
```
so check if works for you and then take that variable approach -- would keep us away from that super long line

yarikoptic:
@nno -- does my hint above works?

nno:
@yarikoptic thanks for the idea. I used a slightly different approach by just adding a new top-level variable. Assuming that travis is happy this can probably be merged then.

yarikoptic:
sure we can merge then... I would have not though indented with tabs there and placed all args for rsync on one line for neater appearance ;)

nno:
You mean no tabs when defining the top-level variable? I would be happy to replace them by spaces.

nno:
@yarikoptic note that the current travis tests fail due to missing mdp as introduced by #435. Should we wait for merging?

yarikoptic:
all checks pass now - I have merged #435 with my workaround etc
merge whenever you feel it looks good ;)

nno:
Done, I will go ahead and merge. Thanks @yarikoptic!

PR #435, opened 2016-03-07
BF: disable testing against MDP master for now, compatibility with recent swig 3.x

yarikoptic:
Needs testing!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,8 @@ before_install:
     # Optional Python modules
     # Note: it looks like the line could be truncated, thus spreading them through multiple lines
     - travis_retry sudo apt-get install shogun-python-modular $PY-psutil $PY-sklearn
-    - travis_retry sudo pip install -e git://github.com/mdp-toolkit/mdp-toolkit#egg=MDP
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -33,7 +33,8 @@ before_install:
     # Optional Python modules
     # Note: it looks like the line could be truncated, thus spreading them through multiple lines
     - travis_retry sudo apt-get install shogun-python-modular $PY-psutil $PY-sklearn
-    - travis_retry sudo pip install -e git://github.com/mdp-toolkit/mdp-toolkit#egg=MDP
+    # Current MDP master after "native 3.x compatibility" RF causes failures, so to be enabled in another PR
```
nno: Disabling MDP causes a doc-test to fail in ``mvpa2/mappers/mdp_adaptor.py``. Should we disable/remove that doctest as well then?

yarikoptic: would be more pain than gain... meanwhile trying to fix up mdp:
https://github.com/mdp-toolkit/mdp-toolkit/pull/16

yarikoptic:
ok -- I am merging this one since fixes up issue with MDP and then I will revert travis to use stock mdp whenever they address it

PR #437, opened 2016-03-08
BF(workaround): numpy 1:1.11.0~b3-1+b1 manages to get corrcoef > 1

yarikoptic:
just a tiny bit, but still:

```
> python -c 'import numpy as np; d = np.random.normal(size=(100, 100)).astype(np.float64); print (np.max(np.corrcoef(d)) - 1.0)'
2.22044604925e-16
```
so causes tests to fail.  yet to check if master numpy fixed


hanke:
Sad that this is necessary... But thanks for implementing it.

PR #440, opened 2016-03-11
BF: deal with various deprecations and their warnings

yarikoptic:


yarikoptic:
everyone loves it (by default) thus merging

PR #441, opened 2016-03-12
ENH: skip testing AUC on KNN based classifiers 

yarikoptic:
since we ruined our ad-hoc support for it in adf13478a4628c0f04b63eabd1cb405d45175a7d


yarikoptic:
ok -- no objections thus merging

PR #444, opened 2016-03-14
BF: free up svm model's vectors correctly (Closes #443)

### LINKED ISSUE #443: Memory consumption by SVM
Dear PyMVPA team,

I suspect a memory leak in the SVM wrappers. Consider this example:

```
import mvpa2.datasets.base
import mvpa2.clfs.svm
import numpy as np

ds = mvpa2.clfs.base.Dataset.from_wizard(samples=np.random.uniform(size=[80, 10000]), targets=np.random.uniform(size=[80]))
for _ in range(2000):
    svm = mvpa2.clfs.svm.SVM(svm_impl="EPSILON_SVR", C=1, kernel=mvpa2.kernels.libsvm.LinearLSKernel())
    svm.train(ds)
    svm.predict(ds)
```

I could run this with PyMVPA 2.4.1 on a (Neuro)Debian machine without problems. However, after upgrading the packages `python-mvpa2` and `python-mvpa2-lib` to their most recent versions via apt (`mvpa2.__version__` prints 2.4.2 after), running this code will result in ever increasing memory consumption (and is sometimes stopped by the OS as a last resort). Is this intended behavior or a memory leak?

Kind regards,
Laurens Berthold

yarikoptic:
May be added few not really needed double free ups 'just incase' and a
unittest for the leak.   Thanks Laurens Berthold (@bertholdl) for the very nice report

### DIFF
--- a/mvpa2/tests/test_svm.py
+++ b/mvpa2/tests/test_svm.py
@@ -9,7 +9,7 @@
 """Unit tests for SVM classifier"""
 
 import numpy as np
-
+import gc
 from mvpa2.datasets import dataset_wizard
 
 from mvpa2.testing import *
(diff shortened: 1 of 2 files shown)


mvpa2/tests/test_svm.py
```diff
@@ -187,6 +187,30 @@ def test_C_on_int_dataset(self, clf):
...
+        ds = mvpa2.clfs.base.Dataset.from_wizard(
+            samples=np.arange(400000).reshape([4, 100000]),
+            targets=np.arange(4))
+
+        for _ in range(5):
+            svm = mvpa2.clfs.svm.SVM(svm_impl="EPSILON_SVR", C=1)
+            svm.train(ds)
+            svm.predict(ds)
+            gc.collect()
+            if not _:
```
nno: I find this hard to read, as the underscore is typically used as a throw-away variable. How about: use ``i`` as loop variable, then
```
is_first_iter = (i==0)
if is_first_iter:
    mem0 = get_vmem()
```

yarikoptic: yeah - redid it a bit without using throw away

yarikoptic:
since releasing takes awhile and I am leaving tomorrow I would appreciate quick final ack/nack to have it merged

PR #448, opened 2016-03-24
NF: I/O support for BIDS _events.tsv files

hanke:


mvpa2/tests/test_bids.py
```diff
@@ -0,0 +1,29 @@
...
+    eq_(load_events(StringIO(evtsv)), [])
+    ra = load_events(StringIO(evtsv), as_recarr=True)
+    eq_(len(ra), 0)
+    eq_(ra.dtype.names, ('onset', 'duration'))
+    # now with content to do type checks
+    evtsv = "onset\tduration\ttrial_type\n2\t1.3\tboring\n3.5\t4\texciting"
+    ra = load_events(StringIO(evtsv))
+    eq_(ra,
+        [{'onset': 2.0, 'duration': 1.3, 'trial_type': 'boring'},
+         {'onset': 3.5, 'duration': 4.0, 'trial_type': 'exciting'}])
```
yarikoptic: what is the big plan though?  would this one get a class analogous to OpenFMRIDataset?

hanke: Yes, I think so. But I don't have the juice to do it in one go -- hence little by little.

Issue #447, opened 2016-03-23
Searchlight with SurfaceQueryEngine and add_center_fa=True mislabels roi_seed


swaroopgj:
@nno @yarikoptic 
`Searchlight` when used with 
1) `SurfaceQueryEngine` 
2) `add_center_fa`=True, and 
3) on a surface data with some nodes masked out

will result in wrong `roi_seed`. 
The issue is that surface query engines have a mapping `_vertex2feature_map` that takes care of masked dataset features, but `Searchlight` doesn't know about it and marks the pre-masked feature index as center node which is might not exist in the masked neighborhood returned by the `SurfaceQueryEngine`. 
As a temporary fix, in my local repo I updated the `SurfaceQueryEngine` to return the neighborhood nodes sorted by the distance to the center, that way keeping the first feature as the center_node.

nno:
Thanks @swaroopgj, I will look into this.

PR #450, opened 2016-04-02
NF: Adds SurfaceRingQueryEngine for surfaces similar to HollowSphere

swaroopgj:
Adds SurfaceRingQueryEngine for surfaces similar to HollowSphere,
with tests and some PEP8 formatting for test_surfing.
@nno @yarikoptic

### DIFF
--- a/mvpa2/misc/surfing/queryengine.py
+++ b/mvpa2/misc/surfing/queryengine.py
@@ -21,7 +21,7 @@ __docformat__ = 'restructuredtext'
 
 import numpy as np
 
-from mvpa2.base.dochelpers import _repr_attrs
+from mvpa2.base.dochelpers import _repr_attrs, borrowkwargs
 
 from mvpa2.base.dataset import AttrDataset
 from mvpa2.misc.neighborhood import QueryEngineInterface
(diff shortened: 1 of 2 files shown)


yarikoptic:
Hey @nno -- could you remind me why I haven't insisted on having similar to volumetric QueryEngines to have clean separation of "geometry" (Sphere, HollowSphere) from the QE itself?

nno:
@yarikoptic I don't know why you did not insist on that. Possibly because "geometry" objects for surfaces have to contain the Surface objects themselves anyway. Alternatively, with only one geometry at the time (standard circle on the surface) this would have seemed overkill back then.

mvpa2/misc/surfing/queryengine.py
```diff
@@ -188,6 +188,111 @@ def query_byid(self, vertex_id):
...
+        v2f = self._vertex2feature_map
+        # Sorting nodes based on distance to center node to work around
+        # the problem with add_center_fa in Searchlight
+        nearby_nodes_keys = sorted(nearby_nodes, key=nearby_nodes.__getitem__)
+        if self.include_center and vertex_id in nearby_nodes_keys:
+            return sum((v2f[node] for node in nearby_nodes_keys
+                        if nearby_nodes[node] > self.inner_radius), v2f[vertex_id])
+        else:
+            return sum((v2f[node] for node in nearby_nodes
+                        if nearby_nodes[node] > self.inner_radius), [])
```
nno: there seems some code duplication here for lines 290 and 293.

nno:
@swaroopgj nice work. Would your approach to sorting the nodes as done here also fix issue #447?

swaroopgj:
Not directly, it could still provide wrong ROI_seed, but atleast we can be
sure that the first node in the list is the center or the closest to the
center.

nno:
@swaroopgj looks good. @yarikoptic any objections to merging?

PR #451, opened 2016-04-04
ENH: allow to drop sa/fa which differ across datasets in hstack/vstack

yarikoptic:
I think we might be brewing incorrect results in some cases where people do not really follow what is happening with their datasets during hstack/vstack'ing.  Paranoid me simply wanted to drop those fa/sa which are not identical (per each feature/sample) across datasets we are stacking, and imho this should have been our default behavior!  I left behavior as is ('update' strategy) for compatibility, but wouldn't mind to "fix" that ;)

unittest is somewhat rudimentary and could be improved but should be sufficient to cover all new code ;)


yarikoptic:
ha -- already saved my butt, (had Fine and FINE as labels, who knows when later would have discovered), although figuring out why it was dropped took awhile ;)

yarikoptic:
@hanke, @nno -- any feedback? ;-)  shouldn't we by default `drop_nonunique` to be on a safer side?

nno:
@yarikoptic actually I would prefer a default value of ``unique`` - in that case an exception is raised by default. It seems the safest default option.

yarikoptic:
@nno `unique` is NotImplemented per this PR ;) I could argue that dropping nonunique would be as safe is not safer ;)

yarikoptic:
@nno  or could argue that dropping is more useful.  Most often some attributes will differ, and I see myself/ppl simple more frequently using `drop_nonunique` instead of default `unique`.   Meanwhile will push a little change making `drop_nonunique` default.

yarikoptic:
@nno I have pushed a RF to make `drop_nonunique` a default but it lead to some other functionality you have introduced (didn't git blame but I think so) to be broken (according to the tests), would you be kind to review those tests etc... may be I am doomed to implement 'unique' logic? ;)
```
ERROR: mvpa2.tests.test_datasetng.test_unique_stack
  File "/usr/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/yoh/proj/pymvpa/pymvpa/mvpa2/tests/test_datasetng.py", line 509, in test_unique_stack
    sa_stack = stack_by_unique_sample_attribute(data, 'x')
  File "/home/yoh/proj/pymvpa/pymvpa/mvpa2/base/dataset.py", line 1083, in stack_by_unique_sample_attribute
    stacked_ds.sa.pop(sa_label)
KeyError: 'x'

FAIL: mvpa2.tests.test_datasetng.test_mergeds
  File "/usr/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/yoh/proj/pymvpa/pymvpa/mvpa2/tests/test_datasetng.py", line 388, in test_mergeds
    assert_true('test' in merged.fa)
AssertionError: False is not true
```

nno:
@yarikoptic Indeed I worked on those functions.

- `mvpa2.tests.test_datasetng.test_unique_stack`: I think `stack_by_unique_sample_attribute` can be removed. Reading it back now I'm not even sure what a use case is.
- ` mvpa2.tests.test_datasetng.test_mergeds`: this test now fails as expected, since the default behaviour now removes the non-matching feature attribute `test`. 

If you want I can update the tests to reflect these ideas.

nno:
@yarikoptic I just sent a PR at https://github.com/yarikoptic/PyMVPA/pull/5 with updated tests

yarikoptic:
On Mon, 18 Apr 2016, Nikolaas N. Oosterhof wrote:

>    @yarikoptic Indeed I worked on those functions.

>      * mvpa2.tests.test_datasetng.test_unique_stack: I think
>        stack_by_unique_sample_attribute can be removed. Reading it back now
>        I'm not even sure what a use case is.
>      * mvpa2.tests.test_datasetng.test_mergeds: this test now fails as
>        expected, since the default behaviour now removes the non-matching
>        feature attribute test.

>    If you want I can update the tests to reflect these ideas.

that would be great!  if to remove anything, better goes as separate PR
(if not too much "to fix" to be removed later on)

-- 
Yaroslav O. Halchenko

yarikoptic:
@hanke as fix in dadc208 has revealed, some code relies on old default logic of "update" while h/v stacking datasets.  I see more uses ... may be you remember if anything else relied on ```update``` logic?
```
$> git grep '[hv]stack(' | grep -v 'np\..stack' | grep -v 'tests/'
doc/examples/hyperalignment.py:wsc_results = vstack(wsc_results)
doc/examples/hyperalignment.py:ds_mni = vstack(ds_all)
doc/examples/hyperalignment.py:    ds_hyper = vstack(ds_hyper)
doc/examples/hyperalignment.py:bsc_hyper_results = hstack(bsc_hyper_results)
doc/examples/hyperalignment.py:fscores = np.mean(np.asarray(vstack(fscores)), axis=0)
doc/examples/hyperalignment.py:ds_hyper = vstack(ds_hyper)
doc/examples/hyperalignment.py:ds_fs = vstack(ds_fs)
doc/examples/hyperalignment.py:        ds_hyper = vstack(ds_hyper)
doc/examples/mdp_mnist.py:ds = vstack((train, test))
doc/examples/pylab_2d.py:ds_lin = vstack((patternsPos, patternsNeg))
doc/examples/smlr.py:trainpat = vstack((patternsPos, patternsNeg))
doc/examples/smlr.py:testpat = vstack((patternsPos, patternsNeg))
doc/source/tutorial_mappers.rst:>>> # this is PyMVPA's vstack() for merging samples from multiple datasets
doc/source/tutorial_mappers.rst:>>> fds = vstack(run_datasets, a=0)
mvpa2/algorithms/benchmarks/hyperalignment.py:        ds_test = vstack(dss_test_bc)
mvpa2/base/dataset.py:                "Instead of ds.append(x) use: ds = vstack((ds, x), a=0)")
mvpa2/base/dataset.py:def vstack(datasets, a=None, fa='drop_nonunique'):
mvpa2/base/dataset.py:def hstack(datasets, a=None, sa='drop_nonunique'):
mvpa2/base/dataset.py:    stacked_ds = hstack(ds, True)
mvpa2/base/dataset.py:    stacked_ds = vstack(ds, True)
mvpa2/cmdline/cmd_crossval.py:    ds = vstack(dss)
mvpa2/cmdline/cmd_select.py:    ds = vstack(dss)
mvpa2/cmdline/helpers.py:    return vstack(hdf2ds(sources))
mvpa2/datasets/niml.py:def hstack(dsets, pad_to_feature_index=None, hstack_method='drop_nonunique',
mvpa2/datasets/niml.py:            padded_dset = dataset.hstack((stripped_dset, empty_dset), hstack_method)
mvpa2/datasets/niml.py:    hstack_dset = dataset.hstack(padded_dsets, hstack_method)
mvpa2/datasets/sources/openfmri.py:            dss = vstack(dss, a=0)
mvpa2/featsel/base.py:        return vstack(sliced_ds, True)
mvpa2/featsel/base.py:        return vstack(sliced_data)
mvpa2/mappers/filters.py:            mds = vstack(dses)
mvpa2/measures/anova.py:        results = vstack(results, fa='update')
mvpa2/measures/base.py:            results = vstack(results, True)
mvpa2/measures/base.py:            results = hstack(results, True)
mvpa2/measures/base.py:            sensitivities = vstack(smerged)
mvpa2/measures/searchlight.py:        result_ds = hstack(results)
mvpa2/misc/data_generators.py:    return vstack(dss)
```

hanke:
@yarikoptic I cannot remember details, but given how long the old default behavior has been in place, IMHO it seems the only safe way is to make its use explicit in the present code base. So internal use should remain on whatever level of correctness it was operating on.

yarikoptic:
I thought I have posted a reply but I guess it "didn't go through".  Ok -- I will go through those internal uses of `[hv]stack`, BUT IMHO it would be valuable to fix "level of correctness" whenever it was operating IMHO incorrectly, i.e. possibly introducing misleading fa/sa's into the resultant dataset.  I know that it was a "documented" feature but that documentation was only on the level of hstack/vstack, so whatever functionality was using them internally -- it could be only a guesswork to figure out what has been actually done to get those datasets folks got outside.

yarikoptic:
oh... that was april when we last talked about this PR... @hanke -- any comment on my last comment?  not sure if I would have juice to go through all the internal invocations and try to figure out the intent.  IMHO dropping nonunique values is correct safe logic  to follow to avoid allowing to shoot yourself in a foot.  Since tests pass, I guess no conscious intent was behind those internal uses to retain only the first entry, so I would vote to merge and fix up if any specific usecase shows me wrong somehow

hanke:
I cannot say that I grasp all potential consequences at this point, but I am equipped with an enormous amount of trust ;-)

Please go ahead and merge, but please also make sure to add a significant changelog entry that points this potential change of behavior out.

yarikoptic:
@hanke , smth like 6b26dd07880276c7e24b82c56ecbda9907a5a714 ? ;)

PR #452, opened 2016-04-05
BF: Fixing hyperalignment example so that it doesn't crash without lapack

swaroopgj:
Using alpha values for regularized hyperalignment that don't fail for numpy SVD implementation.


yarikoptic:
hm, never failed for us before I believe.  Under which numpy do they fail for you @swaroopgj ?

swaroopgj:
@yarikoptic 
numpy v 1.9.3 on OS X
I used this data linked below. Could that be the issue?
http://data.pymvpa.org/datasets/hyperalignment_tutorial_data/hyperalignment_tutorial_data_2.4.hdf5.gz

yarikoptic:
Let it be -- examples must be stable! ;)

PR #453, opened 2016-04-08
enable custom cluster based thresholding for non-volumic dataset

bpinsard:
modified the GroupClusterThreshold to enable clustering using custom neighborhood definition (as a sparse matrix of connectivity) that can process surface based dataset for example.

from an array ```tris``` defining the triangles of the surface, connectivity can be obtained with:
```
scipy.sparse.coo_matrix((
            np.ones(3*tris.shape[0]),
            (np.hstack([tris[:,:2].T.ravel(),tris[:,1]]),
             np.hstack([tris[:,1:].T.ravel(),tris[:,2]]))))
```

I prefer to provide neighborhood than do something surface-specific as I have dataset mixing surfaces and ROIs (HCP-style).

Runs quite fast (comparable to the volume based).

Testing needed, but wanted to have feedback on the structure.

### DIFF
--- a/mvpa2/algorithms/group_clusterthr.py
+++ b/mvpa2/algorithms/group_clusterthr.py
@@ -22,7 +22,7 @@ from collections import Counter
 import numpy as np
 
 from scipy.ndimage import measurements
-from scipy.sparse import dok_matrix
+from scipy.sparse import dok_matrix, csgraph, coo_matrix
 
 from mvpa2.mappers.base import IdentityMapper, _verified_reverse1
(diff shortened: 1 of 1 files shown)


yarikoptic:
Please have a look at https://github.com/PyMVPA/PyMVPA/pull/395 which wip to make it work with arbitrary queryengines and extended set of measures... Would be nice to join efforts

bpinsard:
I didn't see that it was WIP, I will follow your pull request. Thanks!

Issue #455, opened 2016-04-10
searchlight hyperalignment tests fails sometimes


nno:
Running travis for PR #454, the following test sometimes fails [1]:
```
======================================================================
FAIL: test_searchlight_hyperalignment (mvpa2.tests.test_searchlight_hyperalignment.SearchlightHyperalignmentTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/nno/PyMVPA/mvpa2/testing/tools.py", line 331, in newfunc
    return func(*arg, **kwargs)
  File "/home/travis/build/nno/PyMVPA/mvpa2/tests/test_searchlight_hyperalignment.py", line 245, in test_searchlight_hyperalignment
    assert(np.sum(max_weight == diag_weight) / float(len(diag_weight)) > 0.98)
AssertionError
```
Ping @swaroopgj 
[1] https://travis-ci.org/nno/PyMVPA/jobs/122044397

swaroopgj:
@nno Thanks Nick. It's an issue with the test being stringent. Will send a PR soon.

nno:
Thanks @swaroopgj, looking forward to the PR (but no rush) :-)

yarikoptic:
Was fixed by #456

nno:
It fails occasionally with
```
assert(np.sum(max_weight == diag_weight) / float(len(diag_weight)) > 0.90)
```
can we use a value lower than 0.90 @swaroopgj?

See: https://travis-ci.org/PyMVPA/PyMVPA/jobs/131643694

swaroopgj:
Thanks Nick. Relaxed the test values for a few tests. Hopefully, this will make them pass for almost all cases.

PR #456, opened 2016-04-10
BF: loosening some stringent tests and some PEP8

swaroopgj:
Addressing issue #455 reported by @nno. 
Current codes needs every feature to have max weight for the same feature in ref_ds in projection matrix, but with certain cases such as when combine_neighborhood=False, or if features are correlated to begin with, it might be too stringent. So, this commit allows for room of at most one feature can be different. @yoh this might be important in the context of daily tests.


swaroopgj:
Integrated these into PR #458

PR #458, opened 2016-04-12
Few bug fixes to handle feature selection, roi_seed, and datasets with different features

swaroopgj:
- Featsel if given as proportion, selects featsel*(nfeatures of ref_ds)
- Adds test to break current version on master without this fix
- Make sure the mapper are intitialized with correct shape
- Make sure roi_seed is not used, and feature corresponding to that seed is selected
when queryengines are surface-to-voxel maps.

@yarikoptic can you please review these changes?

### DIFF
--- a/mvpa2/algorithms/searchlight_hyperalignment.py
+++ b/mvpa2/algorithms/searchlight_hyperalignment.py
@@ -134,8 +134,8 @@ class FeatureSelectionHyperalignment(ClassWithCollections):
             # computing feature scores from the data
             feature_scores = compute_feature_scores(datasets, self.exclude_from_model)
             nfeatures_sel = nfeatures  # default
-            if self.featsel < 1.0 and np.floor(self.featsel * nfeatures):
-                nfeatures_sel = int(np.floor(self.featsel * nfeatures))
+            if self.featsel < 1.0 and int(self.featsel * nfeatures) > 0:
+                nfeatures_sel = int(self.featsel * nfeatures)
             if self.featsel > 1.0:
                 nfeatures_sel = min(nfeatures, self.featsel)
             fselector = FixedNElementTailSelector(nfeatures_sel, tail='upper', mode='select', sort=False)
--- a/mvpa2/algorithms/searchlight_hyperalignment.py
+++ b/mvpa2/algorithms/searchlight_hyperalignment.py
@@ -132,10 +132,12 @@ class FeatureSelectionHyperalignment(ClassWithCollections):
         if self.featsel != 1.0:
             # computing feature scores from the data
             feature_scores = compute_feature_scores(datasets, self.exclude_from_model)
-            if self.featsel < 1.0:
-                fselector = FractionTailSelector(self.featsel, tail='upper', mode='select', sort=False)
-            else:
-                fselector = FixedNElementTailSelector(np.floor(self.featsel), tail='upper', mode='select', sort=False)
+            nfeatures_sel = nfeatures  # default
+            if self.featsel < 1.0 and np.floor(self.featsel * nfeatures):
+                nfeatures_sel = int(np.floor(self.featsel * nfeatures))
+            if self.featsel > 1.0:
+                nfeatures_sel = min(nfeatures, self.featsel)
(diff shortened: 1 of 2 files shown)


mvpa2/algorithms/searchlight_hyperalignment.py
```diff
@@ -132,10 +133,12 @@ def __call__(self, datasets):
         if self.featsel != 1.0:
             # computing feature scores from the data
             feature_scores = compute_feature_scores(datasets, self.exclude_from_model)
-            if self.featsel < 1.0:
-                fselector = FractionTailSelector(self.featsel, tail='upper', mode='select', sort=False)
-            else:
-                fselector = FixedNElementTailSelector(np.floor(self.featsel), tail='upper', mode='select', sort=False)
+            nfeatures_sel = nfeatures  # default
+            if self.featsel < 1.0 and np.floor(self.featsel * nfeatures):
```
nno: Why the floor? If needed, can we use explicitly `np.floor(self.featsel * nfeatures)`?

mvpa2/algorithms/searchlight_hyperalignment.py
```diff
@@ -546,7 +554,12 @@ def __call__(self, datasets):
         # alignment, i.e. the SL ROIs cover roughly the same area
         queryengines = self._get_trained_queryengines(
             datasets, params.queryengine, params.radius, params.ref_ds)
-
+        # For surface nodes to voxels queryengines, roi_seed hardly makes sense
+        if isinstance(queryengines[params.ref_ds], SurfaceVerticesQueryEngine):
+            self.force_roi_seed = False
```
nno: There may be future QueryEngines for which force_roi_seed does, or does not, make sense. Can we make this an attribute or method in the QueryEngine? For example a property  .can_force_roi_seed which is True in the base class, but False in `SurfaceVerticesQueryEngine`. What do you think?

mvpa2/algorithms/searchlight_hyperalignment.py
swaroopgj: Good point. I wrongly assume int() rounds the float, but it does what I want. Will change that to
and int(self.featsel * nfeatures) > 0

mvpa2/algorithms/searchlight_hyperalignment.py
swaroopgj: That will be useful, not only for this but also to fix how `roi_seed` can be wrong for `Searchlight` with `SurfaceVerticesQueryEngine` in the case when `node_id` happens to match with `feature_id`. I will leave it upto the masters to make that decision ;)

nno:
Thanks @swaroopgj. @hanke @yarikoptic any objections against merging?

mvpa2/algorithms/searchlight_hyperalignment.py
nno: Thanks, I've raised a new issue #464.

mvpa2/algorithms/searchlight_hyperalignment.py
```diff
@@ -332,6 +336,8 @@ def __init__(self, **kwargs):
         self.ndatasets = 0
         self.nfeatures = 0
         self.projections = None
+        # This option makes the roi_seed in each SL to be selected during feature selection
+        self.force_roi_seed = True
```
yarikoptic: before next release we should hide all those above as protected (`_force_roi_seed`) or private (`__force_roi_seed`) -- they shouldn't be public modifiable attributes.

mvpa2/tests/test_searchlight_hyperalignment.py
```diff
@@ -153,6 +153,8 @@ def test_hyperalignment_measure(self):
         mappers_fsf_same = fsha_fsf_same(dss_rotated)
         mappers_fsn = fsha_fsn(dss_rotated)
         mappers = fsha(dss_rotated_clean)
+        mappers_diffsizedss = fsha_fsf([sd[:, nfs] for nfs, sd in
```
yarikoptic: I don't see this `mappers_diffsizedss` being checked anywhere below -- so is it just a smoke test which would have led to crash with prev implementation?
NB When I saw "Adds test to break current version on master without this fix" I felt like a dad whose 1st grader brought home his first A ;)

PR #461, opened 2016-04-12
DOC: update matrix plot range to clear confusion, and fixed a few typos

swaroopgj:
Updated the plotting range to [0,2] to reflect the range of correlation distances. This created confusion in one user. Also fixed a few typos while I am at it. @hanke @yarikoptic


yarikoptic:
awesome -- thanks @swaroopgj

Issue #457, opened 2016-04-10
Volume geometry warning for SurfaceVerticesQueryEngine


swaroopgj:
@nno 
`train` spits a warning if volume geometry is not loaded(?) and that includes dumping the dataset onto the console instead of the format used when printing a dataset. 
Is it behavior on purpose? If not, I suggest printing something similar to what print dataset does.

nno:
Adressed by #465

PR #465, opened 2016-04-14
DOC: Changed volume geometry warning to print dataset instead of dumping the dataset

swaroopgj:
Addresses issue #457 
@nno any objections to this change?


nno:
Thanks @swaroopgj, no objections.

PR #468, opened 2016-05-03
ENH+BF(PY3): variety of fixes for PY3 (daily tests), and display also externals version

yarikoptic:


yarikoptic:
ok, should pass now and if does so -- I will merge

PR #469, opened 2016-05-03
WiP:  bogus -- just to check WTF with travis on #468 PR

yarikoptic:


yarikoptic:
hm, so this one built fine :-/

PR #475, opened 2016-05-26
Permit 4D image mask in fmri_dataset if time dimension is 1

effigies:
If you create a mask from a PyMVPA dataset, it will get written with as a 4D image with time dimension of 1. This change makes it possible to use those masks without using nibabel to squeeze out that dimension.


nno:
Thanks for the PR. Could you provide a unit test please?

effigies:
New test passes locally.

mvpa2/tests/test_niftidataset.py
```diff
@@ -442,3 +442,27 @@ def test_strip_nibabel():
...
+    mask4d = pathjoin(pymvpa_dataroot, 'mask4d.nii.gz')
+    # mask4dfail.nii.gz is mask.nii.gz copied twice in the 4th dimension
+    mask4df = pathjoin(pymvpa_dataroot, 'mask4dfail.nii.gz')
+    assert_equal(nibabel.load(mask).shape + (1,),
+                 nibabel.load(mask4d).shape)
+
+    bold1 = fmri_dataset(bold, mask=mask)
+    bold2 = fmri_dataset(bold, mask=mask4d)
+    assert_equal(bold1.shape, bold2.shape)
+    try:
```
nno: How about using ``assert_raises``? here?

effigies:
Fixed.

nno:
Thanks, looks good.

PR #479, opened 2016-06-22
BF: ref_ds indexing issue when used with exclude_from_model

feilong:
When training Hyperalignment, `ref_ds` refers to the index of the reference
Dataset in the training set (after excluding Datasets), which is consistent
with the documentation.
In some other places, the same variable is used as the index of the referece
Dataset in the full set (before excluding Datasets), and that could cause
mismatch issues.

### DIFF
--- a/mvpa2/algorithms/searchlight_hyperalignment.py
+++ b/mvpa2/algorithms/searchlight_hyperalignment.py
@@ -265,9 +265,10 @@ class SearchlightHyperalignment(ClassWithCollections):
     hyperalignment = Parameter(
         Hyperalignment(ref_ds=None),
         doc="""Hyperalignment instance to be used in each searchlight sphere.
-            Default is just the Hyperalignment instance with default parameters.
-            Its `ref_ds` parameter would be overridden and therefore has no effect.
-            """)
+            Default is just the Hyperalignment instance with default
+            parameters. Its `ref_ds` parameter would be overridden by the
+            `ref_ds` parameter of this SearchlightHyperalignment instance
+            because we want to be consistent and only need one `ref_ds`.""")
 
     combine_neighbormappers = Parameter(
         True,
--- a/mvpa2/algorithms/searchlight_hyperalignment.py
+++ b/mvpa2/algorithms/searchlight_hyperalignment.py
@@ -97,13 +97,14 @@ class FeatureSelectionHyperalignment(ClassWithCollections):
     """
 
 
-    def __init__(self, hyperalignment=Hyperalignment(ref_ds=0),
+    def __init__(self, ref_ds=0, hyperalignment=Hyperalignment(ref_ds=0),
                  featsel=1.0, full_matrix=True, use_same_features=False,
                  exclude_from_model=None, dtype='float32', **kwargs):
         """
         For description of parameters see :class:`SearchlightHyperalignment`
         """
         super(FeatureSelectionHyperalignment, self).__init__(**kwargs)
+        self.ref_ds = ref_ds
         self.hyperalignment = hyperalignment
         self.featsel = featsel
         self.use_same_features = use_same_features
@@ -114,7 +115,7 @@ class FeatureSelectionHyperalignment(ClassWithCollections):
         self.dtype = dtype
 
     def __call__(self, datasets):
-        ref_ds = self.hyperalignment.params.ref_ds
+        ref_ds = self.ref_ds
         nsamples, nfeatures = datasets[ref_ds].shape
         if 'roi_seed' in datasets[ref_ds].fa and np.any(datasets[ref_ds].fa['roi_seed']):
             seed_index = np.where(datasets[ref_ds].fa.roi_seed)
@@ -223,7 +224,7 @@ class SearchlightHyperalignment(ClassWithCollections):
     ref_ds = Parameter(0, constraints=EnsureInt() & EnsureRange(min=0),
         doc="""Index of a dataset to use as a reference. First dataset is used
             as default. If you supply exclude_from_model list, you should supply
-            the ref_ds index as index after you remove those excluded datasets.
+            the ref_ds index as index before you remove those excluded datasets.
             Note that unlike regular Hyperalignment, there is no automagic
             choosing of the "best" ref_ds by default.""")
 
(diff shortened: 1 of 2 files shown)


feilong:
@swaroopgj

mvpa2/tests/test_searchlight_hyperalignment.py
```diff
@@ -208,7 +208,7 @@ def test_searchlight_hyperalignment(self):
         for ds in dss[1:]:
             assert_false(np.all(ds_orig.samples == ds.samples))
         # testing checks
-        slhyp = SearchlightHyperalignment(ref_ds=1, exclude_from_model=[1])
+        slhyp = SearchlightHyperalignment(ref_ds=2, exclude_from_model=[1])
```
yarikoptic: But shouldn't `ref_ds=1` be as fine/bad as 2 here?
To at least "smoke test" correct behaviour of not considering excluded dataset I think it would be better to check with "boundary" indexes, e.g. 4 (and ideally both with 0 as well).  This way if there is some incorrect indexing happening we would have a better chance to catch it.   Or test it explicit -- if we replace (after copying) .samples of the excluded dataset, results must not change a bit

feilong: So here is the confusing part: `ref_ds` should be the index after excluding datasets according to the documentation.
There're 3 datasets in this test, and 2 left after excluding one of them. Therefore using `ref_ds=1` wouldn't raise `ValueError` while `ref_ds=2` would (the "boundary" case). Similarly, `ref_ds=0` should always work if there's at least one dataset left for training.

feilong: I think an explicit test is a better option. The model (including feature selection) should be the same, and results of other datasets shouldn't change. However, results of the excluded datasets should change as we're using different .samples for them, right?

yarikoptic: >  > -        slhyp = SearchlightHyperalignment(ref_ds=1, exclude_from_model=[1])
>  > +        slhyp = SearchlightHyperalignment(ref_ds=2, exclude_from_model=[1])
>    So here is the confusing part: ref_ds should be the index after excluding
>    datasets according to the documentation.
>    There're 3 datasets in this test, and 2 left after excluding one of them.
>    Therefore using ref_ds=1 wouldn't raise ValueError while ref_ds=2 would
>    (the "boundary" case). Similarly, ref_ds=0 should always work if there's
>    at least one dataset left for training.
ah right -- I have not looked careful enough and saw nds=5 and thought
that it is the datasets we will use and didn't have a look at the actual
test ;)... So yes -- having ref_ds=2 is a good one to test for exception being thrown ;)

yarikoptic: >    I think an explicit test is a better option. The model (including feature
>    selection) should be the same, and results of other datasets shouldn't
>    change. However, results of the excluded datasets should change as we're
>    using different .samples for them, right?
sounds right! ;)

feilong: Nice, I'll try to make an explicit test.

swaroopgj:
Just seeing this. I think the docstring was a carryover from old version.
Isn't this simpler to update the doc string and just make a change when we initialize Hyperalignment()

feilong:
Right. We could also change the code around Line 500, which would force `hyperalignment.ref_ds` to be the same as `params.ref_ds`. I think it's more natural to use indices of the original datasets.
The only problem is `hyperalignment.ref_ds` and `params.ref_ds` would have different definitions, which could be confusing...

swaroopgj:
I lean towards using params.ref_ds as index in the complete list and updating the ref_ds when initializing Hyperalignment(). It is much less confusing on users IMHO. Docstring has to be updated to reflect that, of course.

feilong:
How about initializing Hyperalignment with `ref_ds=None`, and update it using an index computed from `param.ref_ds` and `exclude_from_model`?
In that case we should also update the docstring of `hyperalignment` parameter and tell users its `ref_ds` would be overridden.

mvpa2/algorithms/searchlight_hyperalignment.py
```diff
@@ -262,9 +263,10 @@ class SearchlightHyperalignment(ClassWithCollections):
             If None, hyperalignment is performed at every voxel (default).""")
 
     hyperalignment = Parameter(
-        Hyperalignment(ref_ds=0),
+        Hyperalignment(ref_ds=None),
         doc="""Hyperalignment instance to be used in each searchlight sphere.
             Default is just the Hyperalignment instance with default parameters.
+            Its `ref_ds` parameter would be overridden and therefore has no effect.
```
yarikoptic: What/how it will be overridden?  I.e. how its value decided -- please describe here, and if @swaroopgj blesses the changes as regarding the functionality, I will merge

feilong: Essentially we just need one `ref_ds` for everything.

We want to use the same `ref_ds` for both the `SearchlightHyperalignment` instance and the `Hyperalignment` instance it's using for hyperaligning local voxels/nodes.

Therefore we just set the `ref_ds` parameter of `Hyperalignment` with the `ref_ds` parameter of `SearchlightHyperalignment` while running, no matter what that parameter was.

yarikoptic: On Tue, 23 Aug 2016, Feilong Ma wrote:

>    In mvpa2/algorithms/searchlight_hyperalignment.py:

>  >          doc="""Hyperalignment instance to be used in each searchlight sphere.
>  >              Default is just the Hyperalignment instance with default parameters.
>  > +            Its `ref_ds` parameter would be overridden and therefore has no effect.

>    Essentially we just need one ref_ds for everything.

>    We want to use the same ref_ds for both the SearchlightHyperalignment
>    instance and the Hyperalignment instance it's using for hyperaligning
>    local voxels/nodes.

>    Therefore we just set the ref_ds parameter of Hyperalignment with the
>    ref_ds parameter of SearchlightHyperalignment while running, no matter
>    what that parameter was.

so say so explicitly here please ;) ("will be overridden with ref_ds of
the SearchlightHyperalignment instance" or alike)

feilong: Great idea!

swaroopgj: If it passes tests, I have no problem merging it.

yarikoptic:
Ok. I will let @Swaroopgj to have a final look, and bless the merge

yarikoptic:
ah since swaroop did bless it in disappeared comment, and tests pass -- merging

feilong:
When the doctests failed, I had no idea why;
Then the doctests passed, and I had no idea why...

PR #415, opened 2016-01-17
improved plotting of flat surfaces

nno:
- Adds support for plotting surfaces where the nodes are almost but not completely on a flat surface, and/or surfaces are rotated (with a different orientation than (0,0,1) being the normal vector).
- More unit tests
- Minor PEP8 improvements

Issue #327, opened 2015-06-05
prep-afni-surf works only with plump datasets


nno:
Using oblique volumes (where the rotation part of the the voxel-to-world mapping matrix has more than 3 non-zero entries) does not give good alignment.

I'm not sure how easy it is to adapt prep-afni-surf to support oblique volumes. 

Alternatively we could raise an error if the volume is oblique. For almost all use cases, de-obliqing the volume (which changes the orientation) should just work and result in properly aligned surfaces.

PR #482, opened 2016-07-11
NF: in pymvpa2-prep-afni-afni, throw an error when experimental volume is oblique

nno:
Oblique volumes do not work well together with AFNI's alignment pipeline

### DIFF
--- a/mvpa2/support/afni/lib_prep_afni_surf.py
+++ b/mvpa2/support/afni/lib_prep_afni_surf.py
@@ -257,9 +257,11 @@ def augmentconfig(c):
             print "Set absolute path for %s: %s" % (pathvar, c[pathvar])
 
     if c['template'] and c['notemplate']:
-        error('Cannot have both template and notemplate')
+        raise ValueError('Cannot have both template and notemplate')
 
     if 'expvol' in c:
+        _ensure_expvol_is_plump(c['expvol'])
+
         p, n, o, e = utils.afni_fileparts(c['expvol'])
 
         if c.get('outvol_space', None) is None:
@@ -288,6 +290,58 @@ def augmentconfig(c):
 
 
     return c
+
+def _ensure_expvol_is_plump(expvol):
+    '''Helper function that raises an error if the experimental volume
+        is oblique, because such volumes are not supported (yet)'''
+    cmd = '3dinfo "%s"' % expvol
+    info = utils.cmd_capture_output(cmd)
+    info_lines=info.split('\n')
+
+    prefix='Data Axes Tilt:  '
+    is_plump=None
+    print info_lines
+    for line in info_lines:
+        sp=line.split(prefix)
+        has_match=len(sp)==2 and sp[0]==''
+        if has_match:
+            if is_plump is not None:
+                raise ValueError('Multiple matches for prefix "%s"' % prefix)
+            value=sp[1].lower()
+            if 'oblique' in value:
+                is_plump=False
+            elif 'plump' in value:
+                is_plump is True
+            else:
+                raise ValueError('Unable to find orientation')
+
+    if is_plump is None:
+        raise ValueError('Did not find orientation using prefix "%s"' % prefix)
+
+    if not is_plump:
+        raise ValueError('The volume "%s" is oblique (not plump), for which '
+                         'alignment is currently not supported by '
+                         'this AFNI-based pipeline. There are '
+                          'two possible avenues to deal with this:\n'
+                          '(1) copy the volume to another file (e.g. '
+                         'using 3dcopy), then destroy '
+                          'the oblique-ity using 3drefit with the '
+                         '"-deoblique" option. Note that using 3drefit '
+                          'with "-deoblique" '
+                         'destroys some header information, so make sure '
+                         'to make a copy of the volume first.\n'
+                         '(2) run 3dresample with the  "-deoblique" option. '
+                         'Unless 3drefit (option 1 above), this may '
+                         'resample the data.\n'
+                         ' If there are multiple '
+                         'files in the same oblique space (such as multiple '
+                         'fMRI runs) which will be used for subsequent '
+                          'analyses, then it is recommended to deoblique '
+                          'them all and in the same way.' % expvol)
+
+
+
+
 
 def getenv():
     '''returns the path environment


nno:
Addresses https://github.com/PyMVPA/PyMVPA/issues/327

mvpa2/support/afni/lib_prep_afni_surf.py
```diff
@@ -289,6 +291,58 @@ def yesno2bool(d, k): # dict, key
...
+                         'destroys some header information, so make sure '
+                         'to make a copy of the volume first.\n'
+                         '(2) run 3dresample with the  "-deoblique" option. '
+                         'Unless 3drefit (option 1 above), this may '
+                         'resample the data.\n'
+                         ' If there are multiple '
+                         'files in the same oblique space (such as multiple '
+                         'fMRI runs) which will be used for subsequent '
+                          'analyses, then it is recommended to deoblique '
+                          'them all and in the same way.' % expvol)
```
yarikoptic: Why such variable indentation? May be starting string on next line this indented less would also lead to fewer lives ;-)

mvpa2/support/afni/lib_prep_afni_surf.py
```diff
@@ -289,6 +291,58 @@ def yesno2bool(d, k): # dict, key
...
+    prefix='Data Axes Tilt:  '
+    is_plump=None
+    print info_lines
+    for line in info_lines:
+        sp=line.split(prefix)
+        has_match=len(sp)==2 and sp[0]==''
+        if has_match:
+            if is_plump is not None:
+                raise ValueError('Multiple matches for prefix "%s"' % prefix)
+            value=sp[1].lower()
```
yarikoptic: Let's stay pep 8 compliant? Spaces

yarikoptic:
Any chance to get new code covered?

nno:
@yarikoptic thanks for the review, I've made the code more PEP8 compliant. Unfortunately it's not easy to setup coverage for this code, as it would require a big test dataset containing the output from FreeSurfer's recon-all.

mvpa2/support/afni/lib_prep_afni_surf.py
```diff
@@ -166,14 +183,13 @@ def augmentconfig(c):
...
-
     c['prefix_sv2anat'] = 'SurfVol2anat'
 
     # update steps
     if c.get('steps', 'all') == 'all':
-        c['steps'] = 'toafni+mapico+moresurfs+skullstrip+align+makespec+makespecboth+makesurfmasks'
-
-
+        c[
+            'steps'] = ('toafni+mapico+moresurfs+skullstrip+'
```
yarikoptic: yikes

BTW -- by PEP8ing I have meant only the code which was touched by the fix... but ok since work is done already ;)

mvpa2/support/afni/lib_prep_afni_surf.py
```diff
@@ -286,19 +311,75 @@ def yesno2bool(d, k): # dict, key
             c['outvol_ext'] = '+' + c['outvol_view']
             c['outvol_fullext'] = c['outvol_ext'] + '.HEAD'
 
-
     return c
 
+
+
+def _ensure_expvol_is_plump(expvol):
```
yarikoptic: to test this function you would need just a single expvol.  Is it possible to generate it or just ship some small one with necessary features within `mvpa2/data`?

mvpa2/support/afni/lib_prep_afni_surf.py
```diff
@@ -286,19 +311,75 @@ def yesno2bool(d, k): # dict, key
...
+    '''Helper function that raises an error if the experimental volume
+        is oblique, because such volumes are not supported (yet)'''
+    cmd = '3dinfo "%s"' % expvol
```
yarikoptic: although it would require afni(3dinfo) so corresponding check for its presence then ideally could just go to externals.py

mvpa2/support/afni/lib_prep_afni_surf.py
```diff
@@ -286,19 +311,75 @@ def yesno2bool(d, k): # dict, key
...
+    info = utils.cmd_capture_output(cmd)
```
yarikoptic: alternatively (and ideally, since then wouldn't even depend on having afni), this function ( cmd_capture_output) could be mocked while could just be mocked while running a test and provide coverage.  We used mock  already in 1 spot, so should be ok to use testing this one as well (just conditional on having mock module).  Should be smth like
```python

def _test_ensure_expvol_is_plump():
    # actual test function
    ret = _ensure_expvol_is_plump(SOMEVOL)
    # do tests

def test_ensure_expvol_is_plump():
     # first lets try to run as is if 3dinfo is available, so we verify 'integration' with existing afni
     if externals.exists('3dinfo'):
          yield _test_ensure_expvol_is_plump
     # now mocked version... if test runs awhile could be in `else:` I guess
     if externals.exists('mock'):
         import mock
         with patch.object(utils, 'cmd_capture_output', return_value=DESIRED_RETURN):
               yield _test_ensure_expvol_is_plump
```

Then this would test it with and/or without afni installed ;)

mvpa2/support/afni/lib_prep_afni_surf.py
```diff
@@ -286,19 +311,75 @@ def yesno2bool(d, k): # dict, key
...
+
 def getenv():
     '''returns the path environment
     As a side effect we ensure to set for FreeSurfer's HOME'''
     env = os.environ
 
     if 'FREESURFER_HOME' not in env:
-        env['FREESURFER_HOME'] = env['HOME'] # FreeSurfer requires this var, even though we don't use it
+        env['FREESURFER_HOME'] = env[
+            'HOME']  # FreeSurfer requires this var,
```
yarikoptic: please don't! ;-)

mvpa2/support/afni/lib_prep_afni_surf.py
nno: @yarikoptic I agree that having tests for ``ensure_expvol_is_plump`` would be useful. However currently coverage for all functionality in ``lib_prep_afni_surf`` is close to (if not exactly) non-existent. Some integration tests would probably be a better approach compared to writing quite a bit of test code (including tests for checking presence of afni, 3dinfo, ...) with a relative small increase in test coverage. Maybe we can open an issue about coverage for this library and in the meantime just merge?

yarikoptic: well... sure we can merge, but it would yet be another "broken window" while awaiting some miracle to happen to fix them all up (e.g. still awaiting for a test for #415) :-/  since function is specific for handling this case, ideally it should have its own (regardless of having or not an overall integration test) unit-test to make testing fast and "encapsulated".

yarikoptic: IMHO better quick small steps than waiting for a hyperdrive ;)

mvpa2/support/afni/lib_prep_afni_surf.py
yarikoptic: this unnecessary new line remains?

nno:
@yarikoptic, I have now included a small test, trying to follow your suggestion. Please let me know if that follows what you had in mind.

mvpa2/tests/test_lib_afni.py
```diff
@@ -0,0 +1,57 @@
...
+@sweepargs(is_plumb=[False,True])
+@with_tempfile('.nii', 'test_plump')
+def test__ensure_expvol_is_plump(filename, is_plumb, mock_3dinfo):
+    if not mock_3dinfo:
+        # AFNI's 3dinfo is required
+        skip_if_no_external('afni-3dinfo')
+
+    data=np.random.normal(size=(2,2,3,3)).astype(np.int16)
+
+    affine=np.eye(4)
```
yarikoptic: just while at it -- PEP8 ing (spaces around = )

mvpa2/tests/test_lib_afni.py
```diff
@@ -0,0 +1,57 @@
...
+        plump_str='Plumb' if is_plumb else "oblique"
+        mocked_info_lines='Data Axes Tilt:  ' + plump_str
+        mocked_3dinfo = lambda: mocked_info_lines
+        func=lambda: lib_prep_afni_surf._ensure_expvol_is_plump(None,
+                                                        mocked_3dinfo)
+    else:
+        func=lambda: lib_prep_afni_surf._ensure_expvol_is_plump(filename)
+
+    if is_plumb:
+
```
yarikoptic: oh you like empty lines, don't you? ;)

mvpa2/tests/test_lib_afni.py
```diff
@@ -0,0 +1,57 @@
...
+from mvpa2.support.afni import lib_prep_afni_surf
+
+@sweepargs(mock_3dinfo=[False,True])
+@sweepargs(is_plumb=[False,True])
+@with_tempfile('.nii', 'test_plump')
+def test__ensure_expvol_is_plump(filename, is_plumb, mock_3dinfo):
+    if not mock_3dinfo:
+        # AFNI's 3dinfo is required
+        skip_if_no_external('afni-3dinfo')
```
yarikoptic: overall -- neat.  Just don't forget to define a test for `afni-3dinfo` within  `mvpa2/base/externals.py` or let me know if you need help with that

yarikoptic: ah - found it -- thanks! sorry for the noise ;)

nno:
thanks @yarikoptic.

PR #483, opened 2016-07-20
ENH: random and equidistant strategies for FactorialPartitioner

yarikoptic:
as desired on the mailing list by Basil

implementation is probably suboptimal but should be functional.  Noone in a sane mind should use equidistant anyways whenever large number of products between partitionings is expected

did some other minor touches around the code


bpinsard:
works perfectly,
many thanks.

Issue #492, opened 2016-08-22
(Quasi)Deterministic Balancer


effigies:
I've realized that the `Balancer` is random with each call through `generate`, which means that on a cross-validated searchlight analysis, each searchlight may have an entirely different training set. It may be (and, in my case, is) preferred for an entire searchlight analysis to be performed with the same (random) trials excluded, and do multiple searchlight analyses in order to get the resampling desired.

As I see it, the advantage to a non-deterministic `Balancer` is that a smooth region of high classification accuracy is now a "stronger" indicator of information than one in which neighboring searchlights share nearly identical feature vectors, but in a way that seems pretty hard to quantify. By pulling making `Balancer` deterministic in an object-local sense, searchlight becomes a static mapping function, which is easier to reason about. (This applies to all repeated measures.)

I believe that, due to the multiprocessing being process- rather than thread-based, this would actually be a fairly simple modification of saving the random state at `Balancer` creation, and resetting it at the beginning of `generate`, but I'd need to do some testing. It could be necessary to lock around `self(ds)` abd setting/saving state before and after.

So before going much further, I'd like to get some opinions: Would PyMVPA add an object-locally deterministic Balancer? And, if so, should that be (a) standard behavior, (b) optional (`__init__` parameter) behavior, or (c) a subclass?

yarikoptic:
THANKS! See/review/feedback #493? ;)

PR #493, opened 2016-08-22
ENH: provide Balancer with rng= argument, provide get_rng helper

yarikoptic:
Or should we name it 'random_state' to more inline with scikit.learn and actual beast it returns, @hanke -- what you think?

Closes: #492

### DIFF
--- a/mvpa2/generators/resampling.py
+++ b/mvpa2/generators/resampling.py
@@ -86,29 +86,27 @@ class Balancer(Node):
         self._attr = attr
         self.count = count
         self._limit = limit
-        self._limit_filter = None
         self._include_offlimit = include_offlimit
         self._apply_selection = apply_selection
         self._rng = rng
-        self.__rng = None  # will be used within _call but could be initialized
-                           # within generate
 
+    def _get_call_kwargs(self, ds):
+        attr, collection = ds.get_attr(self._attr)
+        # _call might need to operate on the dedicated instantiated rng
+        # e.g. if seed int is provided
+        return {
+            'limit_filter': get_limit_filter(self._limit, collection),
+            'rng': get_rng(self._rng)
+        }
 
-    def _call(self, ds):
+    def _call(self, ds, limit_filter=None, rng=None):
         # local binding
         amount = self._amount
         attr, collection = ds.get_attr(self._attr)
 
-        # get filter if not set already (maybe from generate())
-        if self._limit_filter is None:
-            limit_filter = get_limit_filter(self._limit, collection)
-        else:
-            limit_filter = self._limit_filter
-
         # ids of elements that are part of the balanced set
         balanced_set = []
         full_limit_set = []
-        rng = self.__rng or get_rng(self._rng)
 
         # for each chunk in the filter (might be just the selected ones)
         for limit_value in np.unique(limit_filter):
@@ -194,25 +192,13 @@ class Balancer(Node):
                         "This should never happen!")
             return out
 
-
     def generate(self, ds):
         """Generate the desired number of balanced datasets datasets."""
         # figure out filter for all runs at once
-        attr, collection = ds.get_attr(self._attr)
-        self._limit_filter = get_limit_filter(self._limit, collection)
-        try:
-            # _call might need to operate on the dedicated instantiated rng
-            # e.g. if seed int is provided
-            self.__rng = get_rng(self._rng)
-            # permute as often as requested
-            for i in xrange(self.count):
-                yield self(ds)
-        finally:
-            self.__rng = None
-
-        # reset filter to do the right thing upon next call to object
-        self._limit_filter = None
-
+        # permute as often as requested, reusing the same kwargs
+        kwargs = self._get_call_kwargs(ds)
+        for i in xrange(self.count):
+            yield self(ds, _call_kwargs=kwargs)
 
     def __str__(self):
         return _str(self, str(self._amount), n=self._attr, count=self.count,
--- a/mvpa2/generators/resampling.py
+++ b/mvpa2/generators/resampling.py
@@ -90,6 +90,8 @@ class Balancer(Node):
         self._include_offlimit = include_offlimit
         self._apply_selection = apply_selection
         self._rng = rng
+        self.__rng = None  # will be used within _call but could be initialized
+                           # within generate
 
 
     def _call(self, ds):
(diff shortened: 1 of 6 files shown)


effigies:
I don't think you want this as local to `Balancer._call`, as then you'll always get the same balance, even when `Balancer(count=2)`. Setting in `generate` before the `for i in xrange...`

To make sure we don't over-determine, I think there needs to be a test to the effect of:

```Python
b = Balancer(count=3, rng=XXX)
balanced = list(b.generate(ds))
assert not np.array_equal(balanced[0], balanced[1])
assert not np.array_equal(balanced[0], balanced[2])
assert not np.array_equal(balanced[1], balanced[2])
```

Really, what would be ideal is to have a multi-processing test, to make sure that threads/processes aren't interfering with each other's state, but that's hard. For my [own version](https://github.com/PyMVPA/PyMVPA/compare/master...effigies:balancer_local_determinism), I ran a searchlight while logging excluded trials, and made sure they repeated. But, I was only doing it for `count=1` (my use case).

I don't have any opinions about names.

yarikoptic:
Thanks @effigies -- all this `generate`-vs-`call` is somewhat a fragile point here indeed :-/  I think I have done it correctly now.  But then I have realized that probably the same logic should be within AttributePermutator, BUT somehow if I comment out that instantiation of `rng` within its `generate`, bloody thing still passes the test which supposed to check that all permutations are different... so either a test is incorrect, or a I am missing the point.

Also as travis points out `numpy.random.choice` is apparently a new addition :-/  so I guess I would need to RF it to use `np.random.permutation()[:count]` if we decide to stick with using `numpy` here instead of as originally pure Python's `random`.

mvpa2/tests/test_generators.py
```diff
@@ -221,6 +241,24 @@ def test_balancer():
...
+    bal = Balancer(apply_selection=True, count=5, rng=1)
+    assert_false(np.any(bal(ds).sa.ids != bal(ds).sa.ids))
+
+    # But results should differ if we use .generate to produce those multiple
+    # balanced datasets
+    b = Balancer(apply_selection=True, count=3, rng=1)
+    balanced = list(b.generate(ds))
+    assert_false(all(balanced[0].sa.ids == balanced[1].sa.ids))
+    assert_false(all(balanced[0].sa.ids == balanced[2].sa.ids))
+    assert_false(all(balanced[1].sa.ids == balanced[2].sa.ids))
```
effigies: In addition to the sequence not being constant, I would check that the sequence is repeated on a new `generate` call:

```Python
assert_true(all(np.array_equal(ds_a.sa.ids, ds_b.sa.ids)
            for ds_a, ds_b in zip(balanced, b.generate(ds)))
```

effigies:
So the options for `Balancer` and `AttributePermutator` are:

a) `rng=None` => use global random state
b) `rng=RandomState()` => local random state, updates with each call
c) `rng=int()` => local, determined random state, resets with each call

First, it's not clear to me what the use case for (b) is, except, I suppose, it lets you seed your own at the beginning and get deterministic output without repetitive behavior.

My use case would almost always be:

```Python
Balancer(..., rng=np.random.randint(2**32))
```

Basically, I want a constant seed throughout the life of the object, but I don't want to hard-code one, and I really don't care what its value is, except perhaps to log it.

Any chance of making this canonical and adding:

d) `rng='nonce'` => local, randomly determined initial state, resets with each call

```Python
self._rng = np.random.randint(2**32) if rng == 'nonce' else rng
```

Feel free to reject if this is getting too custom. :-) Also, the name `nonce` isn't important, if you have a better.

mvpa2/tests/test_generators.py
yarikoptic: done, pushed. THANKS!

yarikoptic:
yes, use case for b) is quite scarce.  I guess it would be a localized version of mvpa2.seed which resets both np.random and random RNGs at that point, so that entire analysis could be reproduced.  Here it could be used if someone doesn't need reproducible .generate runs, but overall wants Balancer or AttributePermutator produce reproducible results through a sequence of calls.

`nonce` (or worded alike) would be a neat convenience I would say.  But not sure if I would like to jump implementing it right away... It feels that we would like overall overhaul and unification of such per-instance seedings, and if so, we should unify handling of string values passed as well then... would you be kind to file a separate `[wishlist]` issue for that?

effigies:
Sounds good. Will post shortly.

Final review:

- Tests: Look good.
- Docs: I would add a description of the `rng` parameter to the `AttributePermutator` docstring. And it would be good to describe the separate impacts of a constant seed on `__call__` and `generate`.
- Submitted an optional patch which clears up some execution-ordering edge cases.

yarikoptic:
@hanke , @effigies -- what do you think about this PR in current state and 2d82c6c RFing in particular ;)

mvpa2/generators/resampling.py
```diff
@@ -76,31 +76,38 @@ def __init__(self,
...
+    def _get_call_kwargs(self, ds):
+        attr, collection = ds.get_attr(self._attr)
+        # _call might need to operate on the dedicated instantiated rng
+        # e.g. if seed int is provided
+        return {
+            'limit_filter': get_limit_filter(self._limit, collection),
+            'rng': get_rng(self._rng)
+        }
+
+    def _call(self, ds, limit_filter=None, rng=None):
```
effigies: If I'm reading this correctly, `_call` must *always* be called with `rng` (or L150) will fail.

Does it make sense for these to default to `None`, then?

effigies:
This is much better than hidden state. If it passes, then I'm good with it.

mvpa2/generators/resampling.py
yarikoptic: Well, it is more just to make them keyword arguments... There is no way to enforce them being provided though in python 2 afaik... But they always should be provided unless someone provides wrong `_call_kwargs`... So overall I wouldn't sweat about assuring then not being None for this one ... Tests pass ;-)

yarikoptic:
cool, tests pass, so I will just wait for @hanke 's blessing ;)

yarikoptic:
I was even thinking to refactor more to provide "RepeatingNode" subclass of Node which would absorb common logic here for "generate" between Balancer and Permutator but since it was only 2 of them (didn't find any other on a quick look) decided to leave it alone for now at least ;)

yarikoptic:
ok no complaints, merging, release is unavoidable! (after daily even more thorough testing is ran, so in few days)

PR #494, opened 2016-08-23
BF: scatterplot histogram and boxplot reflect masked data

mvdoc:


mvdoc:
hold on for this. still figuring it out where to center the histogram ;p

mvdoc:
ok now works as desired :)

mvpa2/misc/plot/scatter.py
```diff
@@ -392,13 +395,21 @@ def jitter_me(x, w):
     #       actually consider masked-out values
     if ax_hist_x is not None:
         ax_hist_x.xaxis.set_major_formatter(nullfmt)
-        histx = ax_hist_x.hist(x, bins=binsx, facecolor='b')
+        if mask is not None:
+            histx = ax_hist_x.hist(x_masked, bins=binsx, facecolor='b')
+        else:
+            histx = ax_hist_x.hist(x, bins=binsx, facecolor='b')
```
yarikoptic: so it could be a 1 liner
```python
histx = ax_hist_x.hist(x if mask is None else x_masked , bins=binsx, facecolor='b')
```
here and below.  duplication is evil!

yarikoptic: actually since used in multiple locations -- may be better assign to some dedicated var (if we don't have it yet)?
``` python
x_hist, y_hist = x, y  if mask is None else x_masked, y_masked
```
once and then use in the hist/boxplot calls?

mvdoc: nice!!

PR #497, opened 2016-08-24
Ran into the issue that ENET crashed (within R), because target names…

wmpauli:
Ran into the issue that ENET crashed (within R), because target names were strings in my dataset, rather that 0/1. This patch fixed it, but somebody who knows something should probably make sure this is indeed the proper way of dealing with the situation. Specifically, I am pretty sure that this type casting doesn't affect the original data, which is passed as arg to _train, but I am not sure whether this would always be the case, or whether "data" is sometimes passed as reference.

### DIFF
--- a/mvpa2/clfs/enet.py
+++ b/mvpa2/clfs/enet.py
@@ -131,11 +131,9 @@ class ENET(Classifier):
     def _train(self, data):
         """Train the classifier using `data` (`Dataset`).
         """
-        data.targets = self._attrmap.to_numeric(data.sa[self.get_space()].value)
-        if set(data.targets) != set([0, 1]):
-            raise ValueError, \
-                "Regressors for logistic regression should be [0,1]. Got %s" \
-                % (set(data.targets),)
+        # R enet expects targets to be numerical values
+        if data.uniquetargets.dtype.type is np.string_:
+            data.targets = self._attrmap.to_numeric(data.sa[self.get_space()].value)
         targets = data.sa[self.get_space()].value[:, np.newaxis]
         enet_kwargs = {}
         if self.__max_steps is not None:
--- a/mvpa2/clfs/enet.py
+++ b/mvpa2/clfs/enet.py
@@ -131,6 +131,11 @@ class ENET(Classifier):
     def _train(self, data):
         """Train the classifier using `data` (`Dataset`).
         """
+        data.targets = self._attrmap.to_numeric(data.sa[self.get_space()].value)
+        if set(data.targets) != set([0, 1]):
+            raise ValueError, \
+                "Regressors for logistic regression should be [0,1]. Got %s" \
+                % (set(data.targets),)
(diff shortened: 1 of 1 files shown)


yarikoptic:
Thanks! we have neglected glmnet for a while (since similar functionality can now be used from sklearn)... but let's have it fixed.  Besides this there is other issue for me...  could you please run
```shell
nosetests -s -v mvpa2.tests.test_glmnet
```
and also provide your versions of rpy2 and glmnet?

wmpauli:
Hi,

It completely failed of course, because I neglected that enet doesn't only
what I want it to (classification), but also linear regression (as in your
test). So I changed the code to test whether targets are of dtype
np.string_. Tests run fine now. I will change the pull request.

packageVersion("elasticnet")
[1] ‘1.1’

rpy2.__version__
'2.8.2'

Thanks for the tip on using sklearn instead. After all, elastic net still
crashes in R after running for about an hour.

Wolfgang

mvpa2/clfs/enet.py
```diff
@@ -131,6 +131,9 @@ def __repr__(self):
     def _train(self, data):
         """Train the classifier using `data` (`Dataset`).
         """
+        # R enet expects targets to be numerical values
+        if data.uniquetargets.dtype.type is np.string_:
+            data.targets = self._attrmap.to_numeric(data.sa[self.get_space()].value)
```
yarikoptic: note that you are working with `get_space` but then checking and overriding (not good) targets (incorrect)

all of the checks and operations should be defined by self.get_space (look below were targets extracted) and should't do in place modifications (just assign to targets)

unittest to show and test this issue would have been ideal (although I think probably it should already be triggered by some generic test)

yarikoptic: ok, don't worry -- I will attend to it @wmpauli . Thanks for the PR

yarikoptic:
actually ENET is not "positioned" for classification here, but just for doing regression analyses.  You could have used `RegressionAsClassifier` adapter we have but I see that in our warehouse I have commented out

```
## Still fails unittests battery although overhauled otherwise.
## # enet from R via RPy2
## if externals.exists('elasticnet'):
##     from mvpa2.clfs.enet import ENET
##     clfswh += RegressionAsClassifier(ENET(),
##                                      descr="RegressionAsClassifier(ENET())")
##     regrswh += ENET(descr="ENET()")
```
so I guess there were some outstanding issues.  Since you could use scikit-learn ones here, I would just not bother trying to address this one, unless you really need it. Please let me know

wmpauli:
Hi,

OK. Sorry for the confusion. I thought that this was the way to go after
reading the relevant chapter in Hastie's Elements of Statistical Learning.
I will switch to using the scikit-learn implementation of glmet then, as
you suggested. Thanks!

Wolfgang

yarikoptic:
Closing for now

PR #503, opened 2016-09-29
Fix typo: manhatten -> manhattan

mvdoc:


mvpa2/clfs/distance.py
```diff
@@ -39,8 +39,8 @@ def absmin_distance(a, b):
     return max(abs(a-b))
 
 
-def manhatten_distance(a, b):
-    """Return Manhatten distance between a and b
+def manhattan_distance(a, b):
```
yarikoptic: ho ho -- nice catch ;)

but since it is a part of the API, we need to do such change 'gracefully'
```python

@deprecated("use correctly spelled manhattan_distance instead")
def manhatten_distance(a, b):
    return manhattan_distance(a, b):
```

mvdoc: done ;-) do you test deprecation warnings?

mvpa2/clfs/distance.py
```diff
@@ -39,12 +40,17 @@ def absmin_distance(a, b):
...
+    """Return Manhattan distance between a and b
     """
     return sum(abs(a-b))
 
 
+@deprecated("Use correctly spelled manhattan_distance instead")
+def manhatten_distance(a, b):
+    return manhattan_distance(a, b)
```
yarikoptic: and how would we know that old code works and you haven't made a typo here? ;)  just include a single call within a unittest please and compare result for a good measure.  Then coverage would be happy, and I would be thrilled to merge it ;)

PR #504, opened 2016-09-29
ENH: Basic implementation of cross-validation rsm using learner and CV.

swaroopgj:
This implements cross-validated RSA.
Adds CDist as a measure that will eventually be passed to `CrossValidation` to get cvRSA.

With help from @mvdoc @snastase @feilong

### DIFF
--- a/mvpa2/measures/rsa.py
+++ b/mvpa2/measures/rsa.py
@@ -64,14 +64,21 @@ class CDist(Measure):
 
     def _call(self, ds):
         test_ds = self._prepare_ds(ds)
+        if test_ds.nsamples != self._train_ds.nsamples:
+            raise ValueError('Dataset should have same sample size for dissimilarity.'\
+                             'train: %d, test: %d'%(self._train_ds.nsamples, test_ds.nsamples))
         # Call actual distance metric
         distds = cdist(self._train_ds.samples, test_ds.samples,
                        metric=self.params.pairwise_metric,
                        **self.params.pairwise_metric_kwargs)
         # Make target pairs
-        distds = Dataset(samples=distds.ravel()[:, None],
-                         sa={'pairs': list(product(self._train_ds.T,
-                                                   test_ds.T))})
+        sa_dict = dict()
+        for k in self._train_ds.sa:
+            if k in test_ds.sa:
+                sa_dict[k] = list(product(self._train_ds.sa.get(k).value,
+                                                   test_ds.sa.get(k).value))
+
+        distds = Dataset(samples=distds.ravel()[:, None], sa=sa_dict)
         return distds
 
 
--- a/mvpa2/measures/rsa.py
+++ b/mvpa2/measures/rsa.py
@@ -10,17 +10,56 @@
 
 __docformat__ = 'restructuredtext'
 
-from itertools import combinations
+from itertools import combinations, product
 import numpy as np
 from mvpa2.measures.base import Measure
 from mvpa2.datasets.base import Dataset
 from mvpa2.base import externals
 from mvpa2.base.param import Parameter
 from mvpa2.base.constraints import EnsureChoice
+from mvpa2.mappers.fx import mean_group_sample
 
 if externals.exists('scipy', raise_=True):
-    from scipy.spatial.distance import pdist, squareform
+    from scipy.spatial.distance import pdist, squareform, cdist
     from scipy.stats import rankdata, pearsonr
+
+
+class CDist(Measure):
+
+    pairwise_metric = Parameter('correlation', constraints='str', doc="""\
+          Distance metric to use for calculating pairwise vector distances for
+          dissimilarity matrix (DSM).  See scipy.spatial.distance.pdist for
+          all possible metrics.""")
+
+    sattr = Parameter(['targets'], doc="""
+        List of sample attributes whose unique values will be used to identify the
+        samples groups. Typically your category labels or targets.""")
+
+    def __init__(self, **kwargs):
+        Measure.__init__(self, **kwargs)
+        self.train_ds = None
+        self.sattr = self.params.sattr
+
+    def _train(self, ds):
+        if self.sattr is not None:
+            mgs = mean_group_sample(attrs=self.sattr)
+            self.train_ds = mgs(ds)
+        else:
+            self.train_ds = ds.copy(deep=True)
+
+    def __call__(self, ds, **kwargs):
+        if self.sattr is not None:
+            mgs = mean_group_sample(attrs=self.sattr)
+            test_ds = mgs(ds)
+        else:
+            test_ds = ds.copy(deep=True)
+        # Call actual distance metric
+        distds = cdist(self.train_ds.samples, test_ds, metric=self.params.pairwise_metric)
+        # Make target pairs
+        distds = Dataset(samples=distds.ravel()[None, ],
+                         fa={'pairs': list(product(test_ds.UT, test_ds.UT))})
+
+        return distds
 
 
 class PDist(Measure):
--- a/mvpa2/measures/rsa.py
+++ b/mvpa2/measures/rsa.py
@@ -65,8 +65,9 @@ class CDist(Measure):
     def _call(self, ds):
         test_ds = self._prepare_ds(ds)
         if test_ds.nsamples != self._train_ds.nsamples:
-            raise ValueError('Dataset should have same sample size for dissimilarity.'\
-                             'train: %d, test: %d'%(self._train_ds.nsamples, test_ds.nsamples))
+            raise ValueError('Datasets should have same sample size for dissimilarity, '\
+                             'nsamples for train: %d, test: %d'%(self._train_ds.nsamples,
+                                                                 test_ds.nsamples))
         # Call actual distance metric
         distds = cdist(self._train_ds.samples, test_ds.samples,
                        metric=self.params.pairwise_metric,
--- a/mvpa2/measures/rsa.py
+++ b/mvpa2/measures/rsa.py
@@ -68,8 +68,8 @@ class CDist(Measure):
                        metric=self.params.pairwise_metric,
                        **self.params.pairwise_metric_kwargs)
         # Make target pairs
-        distds = Dataset(samples=distds.ravel()[None, ],
-                         fa={'pairs': list(product(self._train_ds.T,
+        distds = Dataset(samples=distds.ravel()[:, None],
+                         sa={'pairs': list(product(self._train_ds.T,
                                                    test_ds.T))})
         return distds
 
--- a/mvpa2/measures/rsa.py
+++ b/mvpa2/measures/rsa.py
@@ -55,7 +55,7 @@ class CDist(Measure):
             mgs = mean_group_sample(attrs=self.params.sattr)
             ds_ = mgs(ds)
         else:
-            ds_ = ds.copy(deep=True)
+            ds_ = ds.copy()
         return ds_
 
     def _train(self, ds):
--- a/mvpa2/measures/rsa.py
+++ b/mvpa2/measures/rsa.py
@@ -25,25 +25,26 @@ if externals.exists('scipy', raise_=True):
 
 
 class CDist(Measure):
-    """Compute dissimiliarity matrix for samples in a dataset
+    """Compute cross-validated dissimiliarity matrix for samples in a dataset
 
     This `Measure` can be trained on part of the dataset (for example,
     a partition) and called on another partition. It can be used in
     cross-validation to generate cross-validated RSA.
+    Returns flattened dissimilarity values.
     """
-    pairwise_metric = Parameter('correlation', constraints='str', doc="""
-          Distance metric to use for calculating pairwise vector distances for
-          dissimilarity matrix (DSM).  See scipy.spatial.distance.cdist for
-          all possible metrics.""")
+    pairwise_metric = Parameter('correlation', constraints='str',
+            doc="""Distance metric to use for calculating pairwise vector distances for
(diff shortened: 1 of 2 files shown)


mvpa2/measures/rsa.py
```diff
@@ -10,19 +10,58 @@
...
 
 if externals.exists('scipy', raise_=True):
-    from scipy.spatial.distance import pdist, squareform
+    from scipy.spatial.distance import pdist, squareform, cdist
     from scipy.stats import rankdata, pearsonr
 
 
+class CDist(Measure):
+
+    pairwise_metric = Parameter('correlation', constraints='str', doc="""\
```
yarikoptic: minor -- no need for trailing `\` here

mvpa2/measures/rsa.py
```diff
@@ -10,19 +10,58 @@
...
+          all possible metrics.""")
+
+    sattr = Parameter(['targets'], doc="""
+        List of sample attributes whose unique values will be used to identify the
+        samples groups. Typically your category labels or targets.""")
+
+    def __init__(self, **kwargs):
+        Measure.__init__(self, **kwargs)
+        self.train_ds = None
+        self.sattr = self.params.sattr
```
yarikoptic: so why bother binding them to the instance if already available in self.params?
hide away self.train_ds -- unlikely should be "public"

mvpa2/measures/rsa.py
```diff
@@ -10,19 +10,58 @@
...
+            self.train_ds = mgs(ds)
+        else:
+            self.train_ds = ds.copy(deep=True)
+
+    def __call__(self, ds, **kwargs):
+        if self.sattr is not None:
+            mgs = mean_group_sample(attrs=self.sattr)
+            test_ds = mgs(ds)
+        else:
+            test_ds = ds.copy(deep=True)
```
yarikoptic: two pieces of identical code/logic with different input/output usually beg for a helper function/method (e.g. `_prep_dataset`), please refactor

mvpa2/measures/rsa.py
```diff
@@ -10,19 +10,58 @@
...
+        self.sattr = self.params.sattr
+
+    def _train(self, ds):
+        if self.sattr is not None:
+            mgs = mean_group_sample(attrs=self.sattr)
+            self.train_ds = mgs(ds)
+        else:
+            self.train_ds = ds.copy(deep=True)
+
+    def __call__(self, ds, **kwargs):
```
yarikoptic: please overload `_call`, not `__call__` and why there are `**kwargs`??

mvpa2/tests/test_rsa.py
```diff
@@ -73,6 +73,23 @@ def test_PDistConsistency():
...
+    targets = np.tile(xrange(3),2)
+    chunks = np.repeat(np.array((0,1)),3)
+    ds = dataset_wizard(samples=data, targets=targets, chunks=chunks)
+    # Some distance metrics
+    euc = pdist(data, 'euclidean')
+    pear = pdist(data, 'correlation')
+    city = pdist(data, 'cityblock')
+    cdist_euc = CDist(pairwise_metric='euclidean')
+    cdist_pear = CDist(pairwise_metric='correlation')
+    cdist_city = CDist(pairwise_metric='cityblock')
```
yarikoptic: why to hardcode into variables, instead of just a loop 
```python
for dist in ('euclidean', 'cor...):
 pd_ = pdist(data, dist)
 cd_= CDist(pariwise_metric=dist)
...
```
?

mvpa2/measures/rsa.py
```diff
@@ -10,19 +10,63 @@
...
+          Distance metric to use for calculating pairwise vector distances for
+          dissimilarity matrix (DSM).  See scipy.spatial.distance.pdist for
+          all possible metrics.""")
+
+    sattr = Parameter(['targets'], doc="""
+        List of sample attributes whose unique values will be used to identify the
+        samples groups. Typically your category labels or targets.""")
+
+    def __init__(self, **kwargs):
+        Measure.__init__(self, **kwargs)
```
mvdoc: should we be using `super` here?

swaroopgj:
@mvdoc took care of your feedback. Updated with his PR.

yarikoptic:
>  +
>  +    def __init__(self, **kwargs):
>  +        Measure.__init__(self, **kwargs)

>    should we be using super here?

ideally - yes ;)
pragmatically I think effect here would be the same since no multiple
inheritance at this point.

-- 
Yaroslav O. Halchenko

swaroopgj:
Seems like the test failed with unrelated error:
File "/home/travis/build/PyMVPA/PyMVPA/mvpa2/tests/test_benchmarks_hyperalignment.py", line 63, in test_timesegments_classification

mvpa2/measures/rsa.py
```diff
@@ -10,19 +10,63 @@
...
+        self._train_ds = self._prepare_ds(ds)
+        self.is_trained = True
+
+    def _call(self, ds):
+        test_ds = self._prepare_ds(ds)
+        # Call actual distance metric
+        distds = cdist(self._train_ds.samples, test_ds,
+                       metric=self.params.pairwise_metric)
+        # Make target pairs
+        distds = Dataset(samples=distds.ravel()[None, ],
```
swaroopgj: Here, we are arranging samples as folds to be consistent with cross-validation as used with classifiers,
and target pairs as features. So, the datase is already 2-D for a given ROI. But when we run this in a `Searchlight` setting, this might become an issue as it `vstack`s datasets along these "target-pairs" dimension and messing with the feature dimension of `Searchlight`.
One option is to keep the result in each ROI, as nFolds-by-1 dataset and having that 1 element contain a array of "target-pairs". 
Any suggestions on how to handle this? @yarikoptic @mvdoc

swaroopgj: So, I decided to return the result as single-feature dataset. This way it will work with `Searchlight` and users can collapse over cvfolds in `sa` if they wish to.

mvdoc: OK for me about returning samples instead of features. I didn't notice that it was the way Andy returned the pairs in PDist already.

We need to fix the test for `test_CDist_cval` with the travis env when ca are enforced. @swaroopgj do you want me to send PR to your branch or can you modify it?

swaroopgj: So, I think that failure is an issue to be addressed in `CrossValidation`. I am about to open an issue in that regard. 

@mvdoc Do you think we should just by-pass it for now in tests?

mvdoc: Let's see what @yarikoptic says first ;-)

mvpa2/tests/test_rsa.py
```diff
@@ -104,6 +105,9 @@ def test_CDist():
                                       scipy_cdist.ravel())
 
 def test_CDist_cval():
+    if _ENFORCE_CA_ENABLED:
+        # skip testing for now, since we are having issue with 'training_stats'
+        return
```
mvdoc: can you use `SkipTest` instead?

swaroopgj: Will do...

swaroopgj:
yay! It passed.

mvdoc:
:+1: @yarikoptic ready for review ;D

mvpa2/measures/rsa.py
```diff
@@ -10,19 +10,70 @@
...
+from mvpa2.mappers.fx import mean_group_sample
 
 if externals.exists('scipy', raise_=True):
-    from scipy.spatial.distance import pdist, squareform
+    from scipy.spatial.distance import pdist, squareform, cdist
     from scipy.stats import rankdata, pearsonr
 
 
+class CDist(Measure):
+    """Compute dissimiliarity matrix for samples in a dataset
```
yarikoptic: isn't critical piece missing -- "**cross-validated** dissimilarity"

mvpa2/measures/rsa.py
```diff
@@ -10,19 +10,70 @@
...
+    a partition) and called on another partition. It can be used in
+    cross-validation to generate cross-validated RSA.
+    """
+    pairwise_metric = Parameter('correlation', constraints='str', doc="""
+          Distance metric to use for calculating pairwise vector distances for
+          dissimilarity matrix (DSM).  See scipy.spatial.distance.cdist for
+          all possible metrics.""")
+
+    pairwise_metric_kwargs = Parameter({}, doc="""
+    kwargs dictionary passed to cdist. For example,
```
yarikoptic: please indent like in other docs

mvpa2/measures/rsa.py
```diff
@@ -10,19 +10,70 @@
...
+    def _call(self, ds):
+        test_ds = self._prepare_ds(ds)
+        # Call actual distance metric
+        distds = cdist(self._train_ds.samples, test_ds.samples,
+                       metric=self.params.pairwise_metric,
+                       **self.params.pairwise_metric_kwargs)
+        # Make target pairs
+        distds = Dataset(samples=distds.ravel()[:, None],
+                         sa={'pairs': list(product(self._train_ds.T,
+                                                   test_ds.T))})
```
yarikoptic: do not use `.T` since you made it configurable -- should be values of `self.params.sattr`
I wondered if you would have got any advantage of using FlattenMapper while originally just making NxN dataset, then Flattening it, which should provide `1 x N**2` and I believe .fa should bear those pairs).  Do you really need it `N**2 x 1`?

mvpa2/tests/test_rsa.py
```diff
@@ -73,6 +76,53 @@ def test_PDistConsistency():
...
+            # Check to make sure the cdist results are close to CDist results
+            assert_array_almost_equal(res.samples.ravel(),
+                                      scipy_cdist.ravel())
+
+def test_CDist_cval():
+    if _ENFORCE_CA_ENABLED:
+        # skip testing for now, since we are having issue with 'training_stats'
+        raise SkipTest("Skipping test to avoid issue with 'training_stats while CA enabled")
+    
+    targets = np.tile(range(3), 2)
```
yarikoptic: ha -- those lines are not covered now... the cons of our oldfashioned travis setup -- I will fix for it

mvpa2/tests/test_rsa.py
```diff
@@ -73,6 +76,53 @@ def test_PDistConsistency():
...
+                        pairwise_metric_kwargs=metric_kwargs,
+                        sattr=sattr)
+
+            assert_true(not pymvpa_cdist.is_trained)
+            pymvpa_cdist.train(train_data)
+            assert_true(pymvpa_cdist.is_trained)
+            res = pymvpa_cdist(test_data)
+            # Check to make sure the cdist results are close to CDist results
+            assert_array_almost_equal(res.samples.ravel(),
+                                      scipy_cdist.ravel())
```
yarikoptic: since to be trained once and then reused on new data -- would have been nice to test that it provides correct result (comparing to independent new train/test cycle)

mvpa2/tests/test_rsa.py
```diff
@@ -73,6 +76,53 @@ def test_PDistConsistency():
...
+    
+    targets = np.tile(range(3), 2)
+    chunks = np.repeat(np.array((0,1)), 3)
+    ds = dataset_wizard(samples=data, targets=targets, chunks=chunks)
+
+    cv = CrossValidation(CDist(),
+                         generator=NFoldPartitioner(),
+                         errorfx=None)
+    res = cv(ds)
+    # Testing to make sure the both folds return same results, as they should
```
yarikoptic: so if there is a bug and it doesn't recompute the measure across folds we will be just fine? ;)

swaroopgj:
Thanks @yarikoptic, working on the review.

mvpa2/measures/rsa.py
swaroopgj: `self.params.sattr` is a list of attributes to run `mean_group_sample` on, so it doesn't fit well here. 
We can either make sattr a single `sa` instead of a list and use it here instead of `.T` and force `mean_group_sample` to always mean over single `sa` and not accept a list. Any suggestion on which is preferable?

mvpa2/measures/rsa.py
```diff
@@ -10,19 +10,70 @@
...
+    def __init__(self, **kwargs):
+        Measure.__init__(self, **kwargs)
+        self._train_ds = None
+
+    def _prepare_ds(self, ds):
+        if self.params.sattr is not None:
+            mgs = mean_group_sample(attrs=self.params.sattr)
+            ds_ = mgs(ds)
+        else:
+            ds_ = ds.copy(deep=True)
```
yarikoptic: Btw why bother deep copying the data??

mvpa2/measures/rsa.py
yarikoptic: Btw, then why not to keep those which you get from mean group sample? Assign to corresponding sa, fa of your matrix.. I wonder though what flatten maker would do

Also add a test where you have a list and see how your code should break now

swaroopgj: We could keep the sa from meang_group_sample if we don't ravel(). Reason I didn't choose flatten mapper is to flatten the matrix into samples/rows. Flatten mapper does it into columns/features (right?).

mvpa2/measures/rsa.py
swaroopgj: No need. Left over from old implementation.

swaroopgj:
@yarikoptic I have addressed all your comments/requests except how to return results dataset. I will do it once we decide on something. Thanks!

swaroopgj:
@yarikoptic Made changes to return dataset as we discussed this morning.

yarikoptic:
It would be nice to get some example/demo on e.g. comparing CV against "in sample" results

PR #507, opened 2016-10-20
NF: Log and apply selections

effigies:
Following on #493, I've found these useful for debugging and verification of Balancer selections.

Example usage:

``` Python
balancer = Balancer(..., apply_selection=False)
logger = LogExclusions(log_file)
partitioner = ChainNode([NFoldPartitioner(), balancer, logger, ApplySelection()],
                        space='partitions')
```

Any interest in including either of these?

### DIFF
--- a/mvpa2/generators/resampling.py
+++ b/mvpa2/generators/resampling.py
@@ -10,6 +10,8 @@
 
 __docformat__ = 'restructuredtext'
 
+import time
+
 import numpy as np
 
 from mvpa2.base.node import Node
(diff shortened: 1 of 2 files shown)


mvpa2/generators/resampling.py
```diff
@@ -203,3 +205,92 @@ def generate(self, ds):
     def __str__(self):
         return _str(self, str(self._amount), n=self._attr, count=self.count,
                     apply_selection=self._apply_selection)
+
+
+class LogExclusions(Node):
+    """Log excluded entries
+
+    Given a dataset with a boolean sample or feature attribute, return a
+    dataset with only those samples/features marked `True`.
```
yarikoptic: sounds very similar to what Splitter does (usually used after applying a Partitioner which would create such as sample attribute), but that one indeed wouldn't work across features

mvpa2/generators/resampling.py
```diff
@@ -203,3 +205,92 @@ def generate(self, ds):
...
+    """
+    def __init__(self, fname, append=True, space='balanced_set', **kwargs):
+        """
+        Parameters
+        ----------
+        space : str
+          Name of the selection marker attribute in the input dataset that
+          indicates the desired subset.
+        """
+        Node.__init__(self, space=space, **kwargs)
```
yarikoptic: best is to use `super` while invoking super class's constructor... we do use multiple inheritance at places

mvpa2/generators/resampling.py
```diff
@@ -203,3 +205,92 @@ def generate(self, ds):
...
+                fobj.write('\n')
+
+        return ds
+
+
+class ApplySelection(Node):
+    """Applies a selection to a dataset
+
+    Given a dataset with a boolean sample or feature attribute, return a
+    dataset with only those samples/features marked `True`.
```
yarikoptic: this 2 doc lines seems to be way to similar to the previous class's ones... so what is the difference?
also ApplySelection sounds way too generic and could be of different kinds

mvpa2/generators/resampling.py
```diff
@@ -203,3 +205,92 @@ def generate(self, ds):
...
+        elif collection is ds.fa:
+            out = ds[:, battr]
+            del out.fa[space]
+        else:
+            # paranoid
+            raise RuntimeError(
+                    "Don't know where this collection comes from. "
+                    "This should never happen!")
+
+        return out
```
yarikoptic: any new code should have unit-tests to know if it works as it was intended ;)

mvpa2/generators/resampling.py
effigies: Yeah, I copied the docstring from this one and didn't fully edit it, it seems.

mvpa2/generators/resampling.py
effigies: Will put together a test when I get a few more minutes.

yarikoptic: would be great to have it ;)

effigies:
Thanks for the reminder. Tests added.

yarikoptic:
woohoo -- soon we could celebrate 2nd anniversary ;) sorry about that @effigies . I think the world will only get better with this piece so let's press merge ;)

PR #509, opened 2016-11-19
ENH: allow scatter to load gzipped hdf5s

mvdoc:


yarikoptic:
Haha thanks... Anyways I think that whoever ads that extension without actually using gzip to compress entire .hdf5 is misleading by adding that extension... And if it is actually gzipped it wouldn't actually work... Meanwhile merging

yarikoptic:
Ah, missed that it wasn't just fixing the typo... So the question is either those are actually gzipped entirely?

yarikoptic:
What is output of running file command on such a fine you have in hands? ;-)

yarikoptic:
Fwiw failing test is unrelated

mvdoc:
wait, why it shouldn't work? I just saved an hdf5 with `compression='gzip` and loaded w/ `h5load` and it works fine

yarikoptic:
That is to use compression internally within hdf5 for data arrays.

mvdoc:
OK then nevermind :-)

PR #510, opened 2017-01-16
BF: Fix a bug with computing edge lengths.

feilong:
Fix a bug where `p` (or `v[a]`) doesn't update in the for loop.
With this bug, the code actually compute distance between nodes `(0, 1)`, `(0, 2)`, and `(0, 0)` of each face, whereas they should be `(0, 1)`, `(1, 2)`, and `(2, 0)`.
The computed distances (edge lengths) would be underestimated because some of the computed distances are always 0.

### DIFF
--- a/mvpa2/support/nibabel/surf.py
+++ b/mvpa2/support/nibabel/surf.py
@@ -155,12 +155,10 @@ class Surface(object):
             sum_dist = np.zeros((n,))
             count_dist = np.zeros((n,))
             a = f[:, 0]
-            p = v[a]
             for j in xrange(3):
                 b = f[:, (j + 1) % 3]
-                q = v[b]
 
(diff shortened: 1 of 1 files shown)


feilong:
@nno @swaroopgj

yarikoptic:
gy -- nice catch @feilong , seems to be proper!  if only we had a unit-test which should have caught this bug earlier!!!  should be easy to craft one

nno:
Thanks @feilong, I will look into it and work on a test.

nno:
@feilong thanks for the fix. I added a test in PR #513 which passes, so I will go ahead and merge.

nno:
To be clear, PR #513 incorporated this PR so @feilong's fix is now merged into master.

feilong:
Thanks @nno !

PR #512, opened 2017-01-31
Minor BFs in Hyperalignment with tests

swaroopgj:
Added checks to make sure the input is a list of datasets. If not, the current version assumes each sample as a dataset and continues as usual. Current version also craps out if only a single dataset is used for training. @yarikoptic @mih

### DIFF
--- a/mvpa2/algorithms/hyperalignment.py
+++ b/mvpa2/algorithms/hyperalignment.py
@@ -198,6 +198,10 @@ class Hyperalignment(ClassWithCollections):
         """
         params = self.params            # for quicker access ;)
         ca = self.ca
+        # Check to make sure we get a list of datasets as input.
+        if not isinstance(datasets, list):
+            raise TypeError("datasets should be a list (of datasets).")
+
         ndatasets = len(datasets)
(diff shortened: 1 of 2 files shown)


mvpa2/algorithms/hyperalignment.py
```diff
@@ -198,6 +198,10 @@ def train(self, datasets):
         """
         params = self.params            # for quicker access ;)
         ca = self.ca
+        # Check to make sure we get a list of datasets as input.
+        if not (isinstance(datasets, list) or isinstance(datasets, tuple)):
```
yarikoptic: 1. could be just `isinstance(datasets, (list, tuple, np.ndarray))` I believe
2. doc says it should be a "sequence", so could be something else than a list or a tuple I guess (e.g. a ndarray) -- something which we could iterate over and which has an order to it (so shouldn't be a set)... list/tuple/ndarray of them indeed. please adjust docstring and use above construct instead of `or`s

swaroopgj: Sounds reasonable. How do I create a np.ndarray() of datasets to test? Couple of ways I tried didn't work. @yarikoptic

PR #514, opened 2017-02-08
NF: Added dimensionality reduction using SVD to Hyperalignment.

swaroopgj:
As originally published in Haxby et al. (2014), dimensionality reduction using SVD is added after Hyperalignment with extra parameter `output_dim` as input. 
If output_dim is specified, HPAL computes projection to SV space derived from training datasets in addition to mappers to the `ref_ds` space.
@yarikoptic @mih

### DIFF
--- a/mvpa2/algorithms/hyperalignment.py
+++ b/mvpa2/algorithms/hyperalignment.py
@@ -190,7 +190,7 @@ class Hyperalignment(ClassWithCollections):
         # Initializing here so that call can access it without passing after train.
         # Moreover, it is similar to commonspace, in that, it is required for mapping
         # new subjects
-        self.svd_mapper = None
+        self._svd_mapper = None
 
 
     @due.dcite(
@@ -301,9 +301,10 @@ class Hyperalignment(ClassWithCollections):
                                             residuals)
         if params.output_dim is not None:
             mappers = self._level3(datasets)
-            self.svd_mapper = SVDMapper()
-            self.svd_mapper.train(self._map_and_mean(datasets, mappers))
-            self.svd_mapper = StaticProjectionMapper(proj=self.svd_mapper.proj[:, :params.output_dim])
+            self._svd_mapper = SVDMapper()
+            self._svd_mapper.train(self._map_and_mean(datasets, mappers))
+            self._svd_mapper = StaticProjectionMapper(
+                proj=self._svd_mapper.proj[:, :params.output_dim])
 
     def __call__(self, datasets):
         """Derive a common feature space from a series of datasets.
@@ -360,9 +361,9 @@ class Hyperalignment(ClassWithCollections):
             else:
                 mappers = [ChainMapper([zm, m]) for zm, m in zip(zmappers, mappers)]
         elif params.alpha < 1:
-                mappers = [ChainMapper([wm, m]) for wm, m in zip(wmappers, mappers)]
+            mappers = [ChainMapper([wm, m]) for wm, m in zip(wmappers, mappers)]
         if params.output_dim is not None:
-            mappers = [ChainMapper([m, self.svd_mapper]) for m in mappers]
+            mappers = [ChainMapper([m, self._svd_mapper]) for m in mappers]
         return mappers
 
 
--- a/mvpa2/algorithms/hyperalignment.py
+++ b/mvpa2/algorithms/hyperalignment.py
@@ -186,7 +186,11 @@ class Hyperalignment(ClassWithCollections):
     def __init__(self, **kwargs):
         ClassWithCollections.__init__(self, **kwargs)
         self.commonspace = None
-        self.svm = None
+        # mapper to a low-dimensional subspace derived using SVD on training data
+        # Initializing here so that call can access it without passing after train.
+        # Moreover, it is similar to commonspace, in that, it is required for mapping
+        # new subjects
+        self.svd_mapper = None
 
 
     @due.dcite(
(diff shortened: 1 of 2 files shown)


swaroopgj:
Travis-CI failed for unrelated test. Any idea why?

FAIL: test_feature_selection_classifier_with_regression (mvpa2.tests.test_clf.ClassifiersTests)
  File "/home/travis/build/PyMVPA/PyMVPA/mvpa2/tests/test_clf.py", line 559, in test_feature_selection_classifier_with_regression
    sample_clf_reg)
AssertionError: Values were set to the predictions in <LARS(model_type='lasso') classifier>.
Ran 639 tests in 570.683s
FAILED (SKIP=22, failures=1)

mvpa2/algorithms/hyperalignment.py
```diff
@@ -179,6 +186,7 @@ class Hyperalignment(ClassWithCollections):
     def __init__(self, **kwargs):
         ClassWithCollections.__init__(self, **kwargs)
         self.commonspace = None
+        self.svm = None
```
yarikoptic: `svm` is a really confusing name to use here ;) moreover would be nice to have a comment what it actually is here

mvpa2/algorithms/hyperalignment.py
```diff
@@ -340,14 +352,15 @@ def __call__(self, datasets):
...
             else:
-                return [ChainMapper([zm, m]) for zm, m in zip(zmappers, mappers)]
-        else:
-            if params.alpha < 1:
-                return [ChainMapper([wm, m]) for wm, m in zip(wmappers, mappers)]
-            else:
-                return mappers
+                mappers = [ChainMapper([zm, m]) for zm, m in zip(zmappers, mappers)]
+        elif params.alpha < 1:
+                mappers = [ChainMapper([wm, m]) for wm, m in zip(wmappers, mappers)]
```
yarikoptic: overindented a bit? ;-)

yarikoptic: overindented a bit?

mvpa2/tests/test_hyperalignment.py
```diff
@@ -171,6 +171,49 @@ def test_basic_functioning(self, ref_ds, zscore_common, zscore_all):
...
+        dss_sv = [svm.forward(sd) for sd in dss_clean_back]
+        # Test for SVD dimensionality reduction
+        for output_dim in range(1, 5):
+            ha = Hyperalignment(output_dim=output_dim)
+            ha.train(dss_rotated_clean)
+            mappers = ha(dss_rotated_clean)
+            dss_clean_back = [m.forward(ds_)
+                              for m, ds_ in zip(mappers, dss_rotated_clean)]
+            for sd in dss_clean_back:
+                assert(sd.nfeatures == output_dim)
```
yarikoptic: here it is now more of a smoke test really... i would have tuned test data do bear needed features -- i.e. generate data only within output_dim, rotate into higher dimensions and see if we recover it back into lower dimensions more or less correctly

mvpa2/algorithms/hyperalignment.py
```diff
@@ -493,3 +506,21 @@ def _level3(self, datasets):
...
+            zscore(ds_, chunks_attr=None)
+            data_mapped[i] = ds_
+        dss_mean = params.combiner2(data_mapped)
+        return dss_mean
+
+    def _reduce_svd(self, dss_mean):
+        zscore(dss_mean, chunks_attr=None)
+        svd = SVDMapper(force_train=True)
+        svd.train(dss_mean)
+        return svd
```
feilong: An OCPD patient pointed out that there is "no newline at end of file."

mvpa2/algorithms/hyperalignment.py
swaroopgj: will add a comment, thanks.

mvpa2/algorithms/hyperalignment.py
swaroopgj: Nope, it is as it should be.

mvpa2/tests/test_hyperalignment.py
swaroopgj: I agree. Currently, it tests if we replicate the intended behavior from the paper.
I will add a test case with noisy features.

mvpa2/algorithms/hyperalignment.py
swaroopgj: 🤓

feilong:
@swaroopgj I could confirm this strange test failure has happened before, it seems it has a certain chance to fail.  You might pass it when you push again.

swaroopgj:
Regarding using smaller dimensional common space, the current `ProcrusteanMapper` already handles it. But I am afraid it's behavior will be different from computing SVD after alignment in high-dimensional space. Something to test for Hyperalignment Benchmarking hackathon ;)

mvpa2/algorithms/hyperalignment.py
swaroopgj: any suggestions for names? :)

mvdoc:
I had a PR on the other private PyMVPA repo for SVD + Hyperalignment. Isn't that what you are referring to @yarikoptic ?

yarikoptic:
@swaroopgj exactly -- it does handle it but IIRC there were no explicit option to trim/reduce dimensionality
@mvdoc yeap -- you had it in your probhyp and indeed it was something a bit different
I just remember that at that point I have compared fancy coreprobhyp one with something much simpler (like I have described) and they were arriving to nearly or the same results.

indeed we could aim to resolve this all during hackathon.  Procrust+SVD solution would also be good, in case if we for some reason decide to swap Procrust with some other mapper, which could not do dim reduction natively.

mvpa2/algorithms/hyperalignment.py
yarikoptic: note that `self` is not used so you could make this a `@staticmethod` or just an independent function helper

swaroopgj:
@yarikoptic Made suggested changes.

mvpa2/algorithms/hyperalignment.py
yarikoptic: no it is not ;-) ... pushed commit with a "fix" and also has hidden away `_svd_mapper` - IMHO no need to expose it until we need to

mvpa2/algorithms/hyperalignment.py
```diff
@@ -360,9 +361,9 @@ def __call__(self, datasets):
             else:
                 mappers = [ChainMapper([zm, m]) for zm, m in zip(zmappers, mappers)]
         elif params.alpha < 1:
-                mappers = [ChainMapper([wm, m]) for wm, m in zip(wmappers, mappers)]
```
swaroopgj: Now I see what you meant in your original comment. I totally missed it. :(

yarikoptic:
tests pass, swaroop is happy, I am happy... merging

PR #515, opened 2017-02-10
NF: Parallelizing last step of Hyperalignment using joblib

swaroopgj:
joblib based parallelization of last stage of Hyperalignment.
@yarikoptic @feilong

### DIFF
--- a/mvpa2/algorithms/hyperalignment.py
+++ b/mvpa2/algorithms/hyperalignment.py
@@ -154,6 +154,10 @@ class Hyperalignment(ClassWithCollections):
                 reference.  If `None`, then the dataset with the maximum
                 number of features is used.""")
 
+    nproc = Parameter(4, constraints=EnsureInt() & EnsureRange(min=1),
+            doc="Number of processes to use to parallelize the last step of"
+                "alignment. Requires `joblib` package.")
+
     zscore_all = Parameter(False, constraints='bool',
             doc="""Flag to Z-score all datasets prior hyperalignment.
             Turn it off if Z-scoring is not desired or was already performed.
@@ -488,26 +492,40 @@ class Hyperalignment(ClassWithCollections):
         # key different from level-2; the common space is uniform
         #temp_commonspace = commonspace
 
-        residuals = None
+        residuals = [None]*len(datasets)
         if self.ca['residual_errors'].enabled:
-            residuals = np.zeros((1, len(datasets)))
-            self.ca.residual_errors = Dataset(samples=residuals)
+            #residuals = np.zeros((1, len(datasets)))
+            residuals = np.zeros(len(datasets))
+            self.ca.residual_errors = Dataset(samples=residuals[None, :])
 
         # start from original input datasets again
-        for i, (m, ds_new) in enumerate(zip(mappers, datasets)):
-            if __debug__:
-                debug('HPAL_', "Level 3: ds #%i" % i)
+        if params.nproc == 1:
+            for i, (m, ds_new) in enumerate(zip(mappers, datasets)):
+                if __debug__:
+                    debug('HPAL_', "Level 3: ds #%i" % i)
+                m = get_trained_mapper(ds_new, self.commonspace, m, residuals[i])
+                '''
+                # retrain mapper on final common space
+                ds_new.sa[m.get_space()] = self.commonspace
+                m.train(ds_new)
+                # remove common space attribute again to save on memory
+                del ds_new.sa[m.get_space()]
 
-            # retrain mapper on final common space
-            ds_new.sa[m.get_space()] = self.commonspace
-            m.train(ds_new)
-            # remove common space attribute again to save on memory
-            del ds_new.sa[m.get_space()]
-
-            if residuals is not None:
-                # obtain final projection
-                data_mapped = m.forward(ds_new.samples)
-                residuals[0, i] = np.linalg.norm(data_mapped - self.commonspace)
+                if residuals is not None:
+                    # obtain final projection
+                    data_mapped = m.forward(ds_new.samples)
+                    residuals[0, i] = np.linalg.norm(data_mapped - self.commonspace)
+                '''
+        else:
+            verbose_level_parallel = 50 \
+                if (__debug__ and 'GCTHR' in debug.active) else 0
+            from joblib import Parallel, delayed
+            mappers = Parallel(n_jobs=params.nproc,
+                     pre_dispatch=params.nproc,
+                     verbose=verbose_level_parallel)(
+                delayed(get_trained_mapper)
+                (ds, self.commonspace, mapper, residual)
+                for ds, mapper, residual in zip(datasets, mappers, residuals))
 
         return mappers
 
@@ -523,3 +541,13 @@ class Hyperalignment(ClassWithCollections):
             data_mapped[i] = ds_
         dss_mean = params.combiner2(data_mapped)
         return dss_mean
+
+def get_trained_mapper(ds, commonspace, mapper, residual):
+    ds.sa[mapper.get_space()] = commonspace
+    mapper.train(ds)
+    # XXX Is this required?
+    del ds.sa[mapper.get_space()]
+    if residual is not None:
+        data_mapped = mapper.forward(ds.samples)
+        residual = np.linalg.norm(data_mapped - commonspace)
+    return mapper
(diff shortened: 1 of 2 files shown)


mvpa2/algorithms/hyperalignment.py
```diff
@@ -487,27 +491,42 @@ def _level3(self, datasets):
...
+                    residuals.append(residual)
+        else:
             if __debug__:
-                debug('HPAL_', "Level 3: ds #%i" % i)
+                debug('HPAL_', "Level 3: Using joblib with %d jobs" % params.nproc)
+            verbose_level_parallel = 20 \
+                if (__debug__ and 'HPAL' in debug.active) else 0
+            from joblib import Parallel, delayed
+            res = Parallel(n_jobs=params.nproc,
+                     pre_dispatch=params.nproc, backend='threading',
```
swaroopgj: I used backend='threading' as joblib sometimes freezes with backend='multiprocessing' on OSX. May be make this as an optional parameter?

feilong: I replicated the issue with my macbook.
The following code works on a linux machine, without setting `backend`.
```python
def forward_mapper(ds, hyper):
    hypermap = hyper([ds])[0]
    aligned = hypermap.forward(ds)
    return aligned

with Parallel(n_jobs=n_jobs) as parallel:
    dss_hyper = parallel(delayed(forward_mapper)(ds, hyper) for ds in dss)
```

mvdoc: it would be nice to have backend automatically set to `'threading'` if os is darwin, otherwise `'multiprocessing'`, unless it's defined by the user

apparently multiprocessing doesn't work with apple...

mvdoc: why setting `pre_dispatch` to the number of cores? maybe `'auto'` or twice the number of cores would avoid waiting time if the number of subjects is greater than the number of cores

swaroopgj: In majority of real (read, my) usecases, job time is typically longer than dispatch time and pre-dispatching more jobs than necessary might hog up memory (I think).

mvdoc: @yarikoptic master?

yarikoptic: Yes! And make a comment with which joblib version and python on osx you have observed it
Cheers

mvpa2/algorithms/hyperalignment.py
```diff
@@ -487,27 +491,42 @@ def _level3(self, datasets):
...
-        residuals = None
-        if self.ca['residual_errors'].enabled:
-            residuals = np.zeros((1, len(datasets)))
-            self.ca.residual_errors = Dataset(samples=residuals)
+        # Checking for joblib, if not, set nproc to 1
+        if params.nproc > 1:
+            from mvpa2.base import externals, warning
+            if not externals.exists('joblib'):
+                warning("Setting nproc > 1 requires joblib package, which"
+                        "does not seem to exist. Setting nproc to 1")
```
yarikoptic: Minor - space before 'does'

swaroopgj:
@yarikoptic Made requested changes. Look it over and let me know.

mvpa2/algorithms/hyperalignment.py
```diff
@@ -154,6 +154,10 @@ class Hyperalignment(ClassWithCollections):
                 reference.  If `None`, then the dataset with the maximum
                 number of features is used.""")
 
+    nproc = Parameter(1, constraints=EnsureInt() & EnsureRange(min=1),
+            doc="""Number of processes to use to parallelize the last step of
+                alignment. Requires joblib package.""")
```
yarikoptic: joblib's n_jobs:
```
n_jobs: int, default: 1
    The maximum number of concurrently running jobs, such as the number
    of Python worker processes when backend="multiprocessing"
    or the size of the thread-pool when backend="threading".
    If -1 all CPUs are used. If 1 is given, no parallel computing code
    is used at all, which is useful for debugging. For n_jobs below -1,
    (n_cpus + 1 + n_jobs) are used. Thus for n_jobs = -2, all
    CPUs but one are used.
```
so -- do not EnsureRange and below check for `nproc != 1`, not `> 1`

mvpa2/algorithms/hyperalignment.py
```diff
@@ -182,6 +186,13 @@ class Hyperalignment(ClassWithCollections):
             updated common space, and is subsequently called again after each
             2nd-level iteration.""")
 
+    joblib_backend = Parameter(None, constraints=EnsureChoice('multiprocessing',
+                                                    'threading') | EnsureNone(),
+            doc="""Backend to use for joblib when using nproc>1.
```
yarikoptic: and here commend `nproc != 1` ... although not sure if we should check silly folks specifying 0 ;)

mvpa2/algorithms/hyperalignment.py
```diff
@@ -487,27 +498,48 @@ def _level3(self, datasets):
...
+                debug('HPAL_', "Level 3: Using joblib with %d jobs" % params.nproc)
+            verbose_level_parallel = 20 \
+                if (__debug__ and 'HPAL' in debug.active) else 0
+            from joblib import Parallel, delayed
+            import sys
+            # joblib's 'multiprocessing' backend has known issues of failure on OSX
+            # Tested with MacOS 10.12.13, python 2.7.13, joblib v0.10.3
+            if params.joblib_backend is None:
+                params.joblib_backend = 'threading' if sys.platform == 'darwin' \
+                                        else'multiprocessing'
```
yarikoptic: missing space before `'multiprocessing'` -- and tests didn't scream "SyntaxError" ? ;-)

swaroopgj: weird, no it didn't.

mvpa2/algorithms/hyperalignment.py
```diff
@@ -487,27 +498,48 @@ def _level3(self, datasets):
...
+            res = Parallel(n_jobs=params.nproc,
+                     pre_dispatch=params.nproc, backend=params.joblib_backend,
+                     verbose=verbose_level_parallel)(
```
yarikoptic: a little chaotic formatting,  better may be
```python
res = Parallel(
        n_jobs=params.nproc,
        pre_dispatch=...
        )(
            delayed(get_trained_mapper)
        )(
            (ds, self.commonspace, mapper, self.ca['residual_errors'].enabled)
            for ds, mapper in zip(datasets, mappers)
        )
```
to make it easier to parse?

mvpa2/algorithms/hyperalignment.py
```diff
@@ -523,3 +555,30 @@ def _map_and_mean(self, datasets, mappers):
...
+    commonspace: ndarray
+        Commonspace data.
+    mapper: Mapper
+        Typically ProcrusteanMapper.
+    compute_residual: bool
+        Whether to compute residuals or not. Default is False and returns None.
+    """
+    ds.sa[mapper.get_space()] = commonspace
+    mapper.train(ds)
+    # XXX Is this required?
```
yarikoptic: just copy original comment to get the answer to your question ;)

mvpa2/tests/test_hyperalignment.py
```diff
@@ -219,8 +219,8 @@ def test_hpal_svd_combo(self):
                            k=0)
             sv_corrs.append(ndcs)
         self.assertTrue(
-            np.all(np.abs(np.array(sv_corrs)) >= 0.99),
-            msg="Hyperalignment with dimensionality reduction should have"
+            np.all(np.abs(np.array(sv_corrs)) >= 0.9),
```
yarikoptic: why lowering "precision" so much?  was it failing often and should not be able to recover signal nearly precisely?

swaroopgj: Yes, for some random test datasets when added extra "bogus" features correlate with original principal dimensions.

swaroopgj: This test also compares deriving SVD on 2 datasets and testing it on all 4 test datasets, so that's probably where the slight deviations are coming in. I typically saw values of 0.98 where it failed. But tried to be lenient and used 0.9. Do you prefer I bump it up to 0.95?

swaroopgj:
@yarikoptic made those changes. Let me know what you think.

mvpa2/tests/test_hyperalignment.py
yarikoptic: I feel bad for asking, but would be nice to have them as stringent as we could afford

swaroopgj: No worries. Bumped it upto 0.95.

PR #522, opened 2017-05-03
BF: patchwork to make pymvpa usable with python3 again

yarikoptic:
needs more work, contributions are welcome -- not sure if I will have more time for it tomorrow.  hopefully didn't break anything (yet)


yarikoptic:
ok -- I think this should work without conflicting with previous behavior, so let's merge it to make it easier to manage to work with #525

PR #524, opened 2017-05-26
Crossnobis

bpinsard:
I developed 2 interfaces for Cross-validated mahalanobis distance analysis for RSA. (see: Reliability of dissimilarity measures for multi-voxel pattern analysis. Walther et al. 2015)
One is a Measure, and the other is an optimised Searchlight.
It is not finished yet, and there lack tests and proper doc, but I wanted to know if there is interest to include this in PyMVPA, and if so what are the best design choice.

For example in the searchlight, I allow to pass a residuals dataset to _sl_call, in order to compute covariance for each split/searchlight (i need to add this in the Measure too) to perform multivariate normalization in an computationally optimised way (computing sparse covariance of whole dataset between features in same searchlights only). But I don't know what would be the best way to pass it trough __call__ of the parent Searchlight class.

Any comment welcome.
Thanks

### DIFF
--- a/mvpa2/measures/rsa.py
+++ b/mvpa2/measures/rsa.py
@@ -669,7 +669,7 @@ class CrossNobisSearchlight(Searchlight):
                 train_ds = dataset_residuals[train_idx.samples.ravel()]
                 for i in range(int(sl_ext_conn.nnz/1e5+1)):
                     slz = slice(i*blocksize,(i+1)*blocksize)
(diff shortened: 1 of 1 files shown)


yarikoptic:
I thought I have left comments but apparently didn't ;)

1. code without tests is (most likely) broken code
2. changes to other classes seems broke existing code

I felt that something like that doesn't require an additional class(es) and could be accomplished within existing constructs... I could be wrong though -- want to see some tests/old code working before paying closer attention. 

Cheers!

yarikoptic:
@bpinsard could you please add at least a single simple integration test so we could see intended usage and help bringing things into a state not making other functionality break etc

PR #527, opened 2017-06-22
ENH: add required packages to setup.py

mvdoc:
Note: it is still not perfect because it still assumes that `numpy` is installed, but it is less trivial than I thought to make it work—especially for the compiling of libsvm.


yarikoptic:
cool, thanks

PR #533, opened 2017-09-06
fix spelling

usmanayubsh:


yarikoptic:
Thank you ;-)

yarikoptic:
I will merge it when get to the laptop, apparently with Travis upgrade our tests preparation is broken so tests fail so can't merge on the phone

PR #539, opened 2017-09-30
BF: do not use io.FileIO for writing, only for reading 

### LINKED ISSUE #538: "Not all bytes written to" in afni_niml_dset write
```
/home/nastase/PyMVPA/mvpa2/support/nibabel/afni_niml.pyc in write(fnout, niml, form, prefunction)
    720         n = f.write(s)
    721     if n != len(s):
--> 722         raise ValueError("Not all bytes written to %s" % fnout)

ValueError: Not all bytes written to /home/nastase/social_actions/fmri/1021_actions/derivatives/pymvpa/search_RDMs_sq_p1_lh_TEST.niml.dset
> /home/nastase/PyMVPA/mvpa2/support/nibabel/afni_niml.py(722)write()
    720         n = f.write(s)
    721     if n != len(s):

ipdb> print n, len(s)
2147479552 2657855236

```
also tried to explicitly open that file in 'b'-inary mode with the same "success":
```
ipdb> print n2, len(s)
2147479552 2657855236
```
my guess (first thought it was about unicode<->str conversion somewhere, but unlikely):
- hitting the 2GB boundary for a 'signed int' somewhere? (would be 2147483648, so the reported `n` is just that - 4096, unlikely coincidence)

yarikoptic:
Closes #538

### DIFF
--- a/mvpa2/misc/io/base.py
+++ b/mvpa2/misc/io/base.py
@@ -11,6 +11,7 @@ disk."""
 
 __docformat__ = 'restructuredtext'
 
+import os
 import numpy as np
 import mvpa2.support.copy as copy
 from mvpa2.base.dochelpers import enhanced_doc_string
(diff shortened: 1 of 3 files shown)


nno:
Thanks a lot for looking into this @yarikoptic! Seems you already found the solution :-)

yarikoptic:
ok, broke nothing according to the tests... merging ;)

PR #541, opened 2017-10-17
NF: mean_tpr and mean_fnr errorfxs (might be a better default)

yarikoptic:
In case of the disbalanced sets of samples, mean of the TPR is a more robust measure, which should be `1/n_classes` by chance (unlike the mean accuracy error which would jump between `1/n_classes` and some number defined by the proportion of samples in the dominating samples class).  I could be wrong, but according to my "notes" from few years back (http://www.pymvpa.org/files/OHBM2015_Halchenko.pdf) some toolkits (PRoNTo, PROBID, TDT) already use it as the main error function to avoid such problems while dealing with disbalanced sets.  In case of the balanced sets, it should produce the same error/accuracy as the mean_accuracy

### DIFF
--- a/mvpa2/misc/errorfx.py
+++ b/mvpa2/misc/errorfx.py
@@ -72,6 +72,45 @@ def mismatch_error(predicted, target):
     """
     return np.sum( np.asanyarray(predicted) != target )
 
+
+def mean_tpr(predicted, target):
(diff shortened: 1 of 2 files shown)


yarikoptic:
ok, no objections and I think I did not screw up, so  -- merging

PR #542, opened 2017-10-18
Few fixes for recent sphinx deprecation etc

yarikoptic:
not sure if I did anything correctly ;) but at least docs build again


yarikoptic:
ok, no objections -- merging

PR #543, opened 2017-10-18
Bf numpy1.13.1

yarikoptic:
-  commit 391acd8463d23b171f7fb32731f6a015de8265f4
  
   BF(thanks stringent numpy):  deal with a composite mask (one per x, one per y) for scatter_plot
    
    before numpy allowed the selection mask being smaller than actual
    array, e.g.
    
    np.arange(5)[np.array([True, False, True])]
    
    which is not entirely kosher.  And in our case if both masks were
    provided it did something not entirely kosher I think
    Now it should join both masks into intersection first

- commit 97141ed3f7a538fef3bfa19c9653e80bdfa8adfe

     BF(RF): == None -> is None, != None -> is not None
    
    seems that numpy 1.13.1 stopped tollerating comparing to None and made
    it uniform comparison per each element... so if we would like to really
    compare if anything was set, use "is" not "==" with None

### DIFF
--- a/mvpa2/base/state.py
+++ b/mvpa2/base/state.py
@@ -486,9 +486,9 @@ class ConditionalAttributesCollection(Collection):
 
         `other` can be a ClassWithCollections object or ConditionalAttributesCollection
         """
-        if enable_ca == None:
+        if enable_ca is None:
             enable_ca = []
-        if disable_ca == None:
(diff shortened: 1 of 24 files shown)


yarikoptic:
ok, no objections -- merging

PR #551, opened 2017-11-01
BF(TST): numpy 1.13.3 (in conda) bugfix -- ceil returns float, needs int

yarikoptic:


nno:
Great, thanks for the fix @yarikoptic.

nno:
Very good @yarikoptic. Unfortunately I don't think I have enough expertise on the condo-forge topic to make any useful contribution there...

yarikoptic:
That would be my first package.
Failures aren't really conda specific per se, primary sure to novelty of dependencies and some oddities. Heh

PR #561, opened 2017-12-17
because chisqprob was deprecated in 0.17.0

usmanayubsh:


mvpa2/misc/stats.py
```diff
@@ -80,7 +80,7 @@ def chisquare(obs, exp='uniform'):
 
     # return chisq and probability (upper tail)
     # taking only the elements with something expected
-    return chisq, st.chisqprob(chisq, np.sum(exp_nonzeros) - 1)
+    return chisq, st.chi2.sf(chisq, np.sum(exp_nonzeros) - 1)
```
yarikoptic: Thanks!
do you know from which version st.chi2 is available? we might need to make it conditional

usmanayubsh: @yarikoptic `chisqprob` was deprecated and `chi2.sf` was made available from 0.17.0 (see the changelog [here](https://github.com/scipy/scipy/blob/2526df72e5d4ca8bad6e2f4b3cbdfbc33e805865/doc/release/0.17.0-notes.rst).

yarikoptic: thanks! actually the chi2.sf exists even in 0.10 so it would be safe just to switch to use it without any version check, so merging.

yarikoptic:
failing test had nothing to do with this change

PR #563, opened 2018-01-13
ENH: Robustify h5save() in the presence of file locks

mih:
h5lib 1.10 and beyond lock a file if any process reads or writes it (complete description is here: https://support.hdfgroup.org/HDF5/docNewFeatures/SWMR/Design-HDF5-FileLocking.pdf).

A typical PyMVPA script often has an `h5save()` call at the end to store results. With this new locking behavior, we make a single attempt to save and the entire script crashes if there was a concurrent access at that particular time. This leads to desperation, anger, and unicorns dying each time is happens.

This PR introduces re-trying to obtain a file handler in case it was reported as locked. The present implementation tries a fixed maximum number of 30 times, one second apart. This is of course arbitrary and will not fix the issue in case of large files being written that are also open for read, etc... However, it works for given a poor NFS server a little wiggle room to report a file as unlocked again.

We could also decide to implement completely blocking behavior instead, but I had no juice to think that through and went for this hack instead.

### DIFF
--- a/mvpa2/base/hdf5.py
+++ b/mvpa2/base/hdf5.py
@@ -906,7 +906,26 @@ def h5save(filename, data, name=None, mode='w', mkdir=True, **kwargs):
         target_dir = osp.dirname(filename)
         if target_dir and not osp.exists(target_dir):
             os.makedirs(target_dir)
-    hdf = h5py.File(filename, mode)
+
+    retries = 30
+    while retries:
+        try:
+            hdf = h5py.File(filename, mode)
+            break
+        except IOError as e:
+            # sadly this test is not reliable
+            #if e.errno == 11:
+            if retries > 1 and 'resource temporarily unavailable' in str(e):
+                # cannot lock the file: possibly concurrent use, but also
+                # seems to happen spuriously with HDF5 1.10
+                # give it a second and try once more
+                import time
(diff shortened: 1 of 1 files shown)


yarikoptic:
I do similar trick in datalad when downloads fail... But I grow the delay exponentially... May be for no good reason. That allowed me to avoid failed crawls, so even though a dumb solution.
The question here though is why there is two processes accessing the same file at the same time? You might end up with very inconsistent results if one is reading before another writes, just by a virtue of some kind of race. Definitely inconsistent if multiple processes try to write into the same file... In what scenario did you encounter it? As a backend to pass searchlight results? Or really saving the results? Is it an issue if a lock not being released quickly enough over NFS?
I am just afraid that this would hide away the real issue which would haunt us down later in our adulthood with unicorns becoming demogorgons

mih:
The issue here is the flaky behavior of hdf5lib 1.10. If you google for it
you can find all kinds of reports (for pytables, for general HPC use...).
This is the behavior that is successfully hidden.

In any case, it is inexceptable to just fail after 60 hours of computing
even if whatever other process placed a (read) lock on a file. This PR
makes it possible to _maintain_ the behavior of pymvpa with hdf5lib 1.8 and
earlier.

PR #565, opened 2018-01-24
Added new classifier ShrinkageLDA to gda.py

Twitterfax:
Dear PyMVPA users,

Since I couldn't find an implementation of "shrinkage LDA" (e.g., Pereira & Botvinick, NeuroImage, 2011) in PyMVPA, I implemented my own based on Gaussian Discriminant Analysis (gda.py). I figured it could be helpful for other users as well?

It is considerably faster than using the scikit-learn implementation through the SKLLearnerAdapter, which I have been using until now:

SKLLearnerAdapter(LinearDiscriminantAnalysis(solver='lsqr', shrinkage='auto'))

It is also a little more flexible because you can choose between the Ledoit-Wolf estimator and the Oracle Approximating Shrinkage.

May it be useful.

Cheers,
Michael

### DIFF
--- a/mvpa2/clfs/gda.py
+++ b/mvpa2/clfs/gda.py
@@ -35,12 +35,12 @@ from mvpa2.base.param import Parameter
 from mvpa2.base.constraints import EnsureChoice
 from mvpa2.base.state import ConditionalAttribute
 #from mvpa2.measures.base import Sensitivity
-
+from sklearn.covariance import LedoitWolf, OAS
 
 if __debug__:
     from mvpa2.base import debug
 
-__all__ = [ "LDA", "QDA" ]
+__all__ = [ "LDA", "QDA", "ShrinkageLDA" ]
 
 class GDA(Classifier):
     """Gaussian Discriminant Analysis -- base for LDA and QDA
@@ -259,3 +259,94 @@ class QDA(GDA):
             dm = data - m
             res.append(b - 0.5 * np.sum(np.dot(dm, covi) * dm, axis=1))
         return np.array(res).T
+
+class ShrinkageLDA(GDA):
+    """LDA with Ledoit-Wolf estimate of covariance matrix
+    """
+
+    __tags__ = GDA.__tags__ + ['linear', 'lda', 'shrinkage']
+
+    def __init__(self, shrinkage_estimator='LedoitWolf', block_size=1000, \
+    **kwargs):
+        super(ShrinkageLDA, self).__init__(**kwargs)
+        self.block_size = block_size
+        if shrinkage_estimator is 'LedoitWolf':
+            self.shrinkage_estimator = LedoitWolf(block_size=block_size)
+        elif shrinkage_estimator is 'OAS':
+            self.shrinkage_estimator = OAS()
+        else:
+            raise ValueError, "Unknown estimator '%s'" % shrinkage_estimator
+
+    def _train(self, dataset):
+        """Train the classifier using `dataset` (`Dataset`).
+        """
+
+        # Adapted from GDA's method _train. Couldn't use that implementation
+        # because it only computes the class-dependent SSCP matrices. The
+        # shrinkage coefficient, however, needs to be computed from the whole
+        # time series (with samples centered on their respective class means)
+        params = self.params
+        targets_sa_name = self.get_space()
+        targets_sa = dataset.sa[targets_sa_name]
+
+        # get the dataset information into easy vars
+        X = dataset.samples
+        labels = targets_sa.value
+        self.ulabels = ulabels = targets_sa.unique
+        nlabels = len(ulabels)
+        label2index = dict((l, il) for il, l in enumerate(ulabels))
+
+        # set the feature dimensions
+        nsamples = len(X)
+        nfeatures = dataset.nfeatures
+
+        self.means = means = \
+                     np.zeros((nlabels, nfeatures))
+        # degenerate dimension are added for easy broadcasting later on
+        # XXX might want to remove -- for now taken from GNB as is
+        self.nsamples_per_class = nsamples_per_class \
+                                  = np.zeros((nlabels, 1))
+        self.cov = cov = \
+                     np.zeros((nlabels, nfeatures, nfeatures))
+
+        # Subtract class means to compute within-class covariance later
+        # better loop than repmat! ;)
+        Xdm = list()
+        for l, il in label2index.iteritems():
+            Xl = X[labels == l]
+            nsamples_per_class[il] = len(Xl)
+            # TODO: degenerate case... no samples for known label for
+            #       some reason?
+            means[il] = np.mean(Xl, axis=0)
+            # Instead of calculating within-class sum of squares, only subtract
+            # the class mean here.
+            Xdm.append(Xl - means[il])
+
+        # Concatenate the centered time series
+        Xdm = np.vstack(Xdm)
+
+        # Store prior probabilities
+        self.priors = self._get_priors(nlabels, nsamples, nsamples_per_class)
+
+        if __debug__ and 'ShrinkageLDA' in debug.active:
+            debug('ShrinkageLDA', "training finished on data.shape=%s " % (X.shape, )
+                  + "min:max(data)=%f:%f" % (np.min(X), np.max(X)))
+
+        nlabels = len(self.ulabels)
+
+        # Use shrinkage estimator to compute covariance matrix
+        self.cov = cov = self.shrinkage_estimator.fit(Xdm).covariance_
+
+        # For now as simple as that -- see notes on top
+        covi = self._inv(cov)
+
+        # Precompute and store the actual separating hyperplane and offset
+        self._w = np.dot(covi, self.means.T)
+        self._b = b = np.zeros((nlabels,))
+        for il in xrange(nlabels):
+            m = self.means[il]
+            b[il] = np.log(self.priors[il]) - 0.5 * np.dot(np.dot(m.T, covi), m)
+
+    def _g_k(self, data):
+        """Return decision function values"""
+        return np.dot(data, self._w) + self._b
--- a/mvpa2/clfs/warehouse.py
+++ b/mvpa2/clfs/warehouse.py
@@ -18,7 +18,7 @@ from mvpa2.clfs.meta import FeatureSelectionClassifier, SplitClassifier, \
      MulticlassClassifier, RegressionAsClassifier
 from mvpa2.clfs.smlr import SMLR
 from mvpa2.clfs.knn import kNN
-from mvpa2.clfs.gda import LDA, QDA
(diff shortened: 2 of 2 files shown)


yarikoptic:
Interested in us guiding you @Twitterfax to polish up the PR (make your code tested, not failing as now, etc), or just want us to do that (might take longer ;) )

Twitterfax:
Please do guide me. It would be much appreciated :)

For example: one thing I wasn't sure about was whether scikit-learn was installed along with PyMVPA by default. If it is not available I'd have to integrate the shrinkage estimators manually. I'm guessing that this could be a problem.

How can I inspect the Travis CI details?

yarikoptic:
Great:
- I will leave comments through out your PR which will be visible in the "Files changed" tab
- Travis CI details -- just click on "Details" by the "continuous-integration/travis-ci/pr" above:
scr
then go to the one failing (red) to see that it is indeed about your code is not being guarded against not having sklearn installed .  I will leave comments pointing to how it should be done

I would also recommend installing codecov browser extension which would help to see which lines of your code are covered by the tests and which not, right in your browser while reviewing differences in that "Files changed" tab: https://github.com/codecov/browser-extension#choose-your-browser-below

mvpa2/clfs/gda.py
```diff
@@ -259,3 +259,94 @@ def _g_k(self, data):
...
         return np.array(res).T
+
+class ShrinkageLDA(GDA):
+    """LDA with Ledoit-Wolf estimate of covariance matrix
+    """
+
+    __tags__ = GDA.__tags__ + ['linear', 'lda', 'shrinkage']
+
+    def __init__(self, shrinkage_estimator='LedoitWolf', block_size=1000, \
+    **kwargs):
```
yarikoptic: shave a look at some other classifiers, e.g. at GDA or others to how to ideally define parameters (via those `Parameter` definitions in the class)

no need for trailing `\` here since it is within `()`

mvpa2/clfs/gda.py
```diff
@@ -35,12 +35,12 @@
 from mvpa2.base.constraints import EnsureChoice
 from mvpa2.base.state import ConditionalAttribute
 #from mvpa2.measures.base import Sensitivity
-
+from sklearn.covariance import LedoitWolf, OAS
```
yarikoptic: since this code depends on sklearn presence it might be better to move it into a separate file (e.g. `shrinklda.py` or alike) which would help to import it conditionally on sklearn presence within the warehouse

mvpa2/clfs/gda.py
```diff
@@ -259,3 +259,94 @@ def _g_k(self, data):
...
+    __tags__ = GDA.__tags__ + ['linear', 'lda', 'shrinkage']
+
+    def __init__(self, shrinkage_estimator='LedoitWolf', block_size=1000, \
+    **kwargs):
+        super(ShrinkageLDA, self).__init__(**kwargs)
+        self.block_size = block_size
+        if shrinkage_estimator is 'LedoitWolf':
+            self.shrinkage_estimator = LedoitWolf(block_size=block_size)
+        elif shrinkage_estimator is 'OAS':
+            self.shrinkage_estimator = OAS()
```
yarikoptic: better delay this instantiation until _train, and do not even assign that instance to any ShrinkageLDA attribute then.
It would allow to change that parameter of a classifier instance and retrain and would keep its `repr` correct

mvpa2/clfs/gda.py
```diff
@@ -259,3 +259,94 @@ def _g_k(self, data):
...
+
+    def _train(self, dataset):
+        """Train the classifier using `dataset` (`Dataset`).
+        """
+
+        # Adapted from GDA's method _train. Couldn't use that implementation
+        # because it only computes the class-dependent SSCP matrices. The
+        # shrinkage coefficient, however, needs to be computed from the whole
+        # time series (with samples centered on their respective class means)
+        params = self.params
```
yarikoptic: here we go -- that is where you would find those `Parameter`s after you move them out to be defined at the class level, e.g. `params.shrinkage_estimator`

mvpa2/clfs/gda.py
```diff
@@ -259,3 +259,94 @@ def _g_k(self, data):
...
+            # the class mean here.
+            Xdm.append(Xl - means[il])
+
+        # Concatenate the centered time series
+        Xdm = np.vstack(Xdm)
+
+        # Store prior probabilities
+        self.priors = self._get_priors(nlabels, nsamples, nsamples_per_class)
+
+        if __debug__ and 'ShrinkageLDA' in debug.active:
```
yarikoptic: you would need to add ShrinkageLDA among debug targets alongside with GDA etc:
```shell
$> git grep GDA -- mvpa2/base/   
mvpa2/base/__init__.py:    debug.register('GDA', "GDA - Gaussian Discriminant Analyses")
```

mvpa2/clfs/warehouse.py
```diff
@@ -18,7 +18,7 @@
      MulticlassClassifier, RegressionAsClassifier
 from mvpa2.clfs.smlr import SMLR
 from mvpa2.clfs.knn import kNN
-from mvpa2.clfs.gda import LDA, QDA
+from mvpa2.clfs.gda import LDA, QDA, ShrinkageLDA
```
yarikoptic: after you move that definition into a dedicated file, you could place the import within `if externals.exists('skl')` below so it would be imported only when sklearn installed

Then add an actual instance of the class, similarly to how it is done for LDA, GDA etc in this file.  This should gets its testing going and we will continue from there

Twitterfax:
Awesome! Thanks for your time and help. I will try to implement the suggested changes.

Twitterfax:
Hi Yaroslav. Thanks again for your feedback and patience. Things look ok now... I suppose(?)

yarikoptic:
Good. Now it would be great to add a dedicated unit test which would compare results against sklearn wrapped implementation you mentioned - should be closer to identical right? I can provide a bit more detail later or you could check out other tests to see how it could be done

Twitterfax:
Ok, I have used test_gnb.py etc. as a template to write the test. Let me know if it needs improvement. How would such a test typically be used?

yarikoptic:
> How would such a test typically be used?

Look into .travis.yml (which is a bit a mouthful) for how we invoke testing using nose.  To test locally just run
```
python -m nose -s -v mvpa2/tests/test_shrinklda.py
```
to your added tests only

Twitterfax:
Travis CI doesn't complain about my coding anymore. I take it as a positive sign ;)

Twitterfax:
Hi Yaroslav,
Is there anything else I can do to improve the code? I made the test a little faster by simulating a smaller (but still p > n) dataset, which made it 20 times faster on my computer.
Cheers,
Michael

mvpa2/tests/test_shrinklda.py
```diff
@@ -0,0 +1,74 @@
...
+class ShrinkageLDATests(unittest.TestCase):
+
+    def test_shrinkage_lda_faster_than_skl(self):
+        # Compare PyMVPA's shrinkage LDA implementation with scikit-learn's
+        # implementation when used through SKLLearnerAdapter. Are prediction
+        # accuracies next to identical? Is the PyMVPA implementation faster?
+
+        # First check if scikit-learn is available at all
+        if not externals.exists('skl') or externals.versions['skl'] < '0.17':
+            self.skipTest('sklearn (0.17 or higher) unavailable')
```
yarikoptic: ha -- never used it this way ... according to docs this should indeed raise an exception. So no need for explicit `else:` block indentation -- could be dedented since would not run if there is old sklearn and this exception is raised.

anyways -- you could just use a line
```
skip_if_no_external('skl', min_version='0.17')

File: mvpa2/tests/test_shrinklda.py [hunk 8]
```diff
@@ -0,0 +1,74 @@
...
+            ds = normal_feature_dataset(perlabel=n_samples_per_class,
+            nlabels=n_classes, nfeatures=n_features, nchunks=n_chunks,
+            nonbogus_features=range(n_classes), snr=snr)
+
+            partitioner = NFoldPartitioner(attr='chunks')
+
+            delta_t = []
+            for clf in [pymvpa_clf, skl_clf]:
+                cv = CrossValidation(clf, partitioner)
+                _ = cv(ds)
```
yarikoptic: well, test could compare those results -- your implementation should be not just faster but also providing a similar (better?) result, right? ;-)  otherwise -- what would be the point of having a faster but notably worse implementation?

Twitterfax: Thanks for the feedback!

1.) Yes, that's a cleaner solution with 'skip_if_no_external'.
2.) I though about this but I didn't know how to define the similarity of results. One idea I had was to show that the cv error of the new LDA implementation is more similar to the cv error from the old LDA version than from SVM classification. This is indeed the case in - say - 95 % of the cases. I figured this would not be acceptable for a test which would probably be run routinely together with a bunch of other tests. You don't want to risk the whole test failing just because you're in the last 5 % (or so) of this little test. Do you have a solution for this?

Otherwise I suggest that I elaborate a bit more on the documentation of the new ShrinkageLDA class to explain what it does. The code is quite simple and with a bit more explanation I could help the user convince himself/herself that what it does is indeed basic LDA with shrinkage on the within-class covariance matrix.

Twitterfax:
Thanks for the feedback!

1.) Yes, this is a much cleaner solution.
2.) I thought about this too and the solution I came up with was to compare the cv error of the new classifier with those of the old implementation and the cv error of SVM classification. A comparison between the two shows that the new results are more similar to the old LDA version than SVM. But only in about 95 % of the cases. For a test that should be run routinely, I figured that this would be too low. You don't want the whole test to fail just because you are in the remaining 5 % (roughly) where the test fails, right? Do you have any ideas to solve this?

Otherwise I suggest that I could improve the documentation of the new class. The code is relatively simple but with a little more detail on what the algorithm does the user could easily convince himself/herself that it indeed does what it's supposed to do to implement LDA with shrinkage on the within-class covariance matrix. In the end, this is the point right? In this way the "validity" of the new algorithm would result directly from its implementation rather than from a comparison with other implementations.

Cheers,
Michael

yarikoptic:
Sorry @Twitterfax again for the delay with the feedback, I will try to do better ;)

` to the old LDA version than SVM` - a bit confused why we would want to compare to SVM here.  But I thought that result should be similar to the one produced by sklearn which you compare in speed in against.

re unstable test(s): we do allow for those which only so some times fail.  It is somewhat of our curse but we do have such and do not fix seed to guarantee that it is always the same data/result.  you could use either
1. `if cfg.getboolean('tests', 'labile', default='yes'):` before the test which might fail at times
2. `@labile(5,1)` (from `mvpa2.testing.tools`) to annotate the entire test so it runs up to 5 times to allow for no more than 1 failure (iirc) -- it would need random data regeneration within and we use it more rarely

PR #580, opened 2018-04-23
Added connectivity hyperalignment to algorithms.

swaroopgj:
As described in this paper
http://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1006120

### DIFF
--- a/mvpa2/algorithms/connectivity_hyperalignment.py
+++ b/mvpa2/algorithms/connectivity_hyperalignment.py
@@ -1,9 +1,4 @@
-# -*- coding: utf-8 -*-
-"""
-Created on Feb 03, 2016
-
-@author: Swaroop Guntupalli
-"""
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil; coding: utf-8 -*-
 # vi: set ft=python sts=4 ts=4 sw=4 et:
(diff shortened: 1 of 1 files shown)


yarikoptic:
pushed just some minor cosmetic changes (not changes to any code).  We would need some test(ing) -- did you have any @swaroopgj ?  e.g. as a part of testing any other (generic) hyperalignment?

@feilong will you be interested to help?

feilong:
@yarikoptic Yes, I am happy to help.
Do we have any existing testing dataset that can be used for CHA?
I was also thinking about writing a tutorial using open data, such that people can try and test CHA themselves.

yarikoptic:
We already have mvpa2/tests/test_hyperalignment.py and mvpa2/tests/test_searchlight_hyperalignment.py which could be used (probably searchlight_hyper one since here also needs QEs) to just go over yet another hyperalignment algorithm, and still be able to reconstruct (more or less) original structure.  The signal in the target dataset I guess should get "spatially colored" (e.g. by applying some single random rotation to the entire featureset first which would introduce "correlations/connectivity") first before introducing per-subject rotations which hyperalignment should figure out.

Sure thing complete demo would be great @feilong (but it doesn't rule out the need for a simpler test)

yarikoptic:
@feilong any luck?

PR #584, opened 2018-05-09
Adds searchlight on the surface mesh - using disks instead of spheres

miykael:
This PR adds the functionality to do searchlight analysis on the **surface mesh**, using a **disk** as searchlight.

I understand that PyMVPA has already some surface based searchlight functions, e.g. [`disc_surface_queryengine`](http://www.pymvpa.org/generated/mvpa2.misc.surfing.queryengine.disc_surface_queryengine.html) as described in [Oosterhof et al. (2011)](http://www.pymvpa.org/references.html#owd-11), but as far as I understand, none of them runs the serachlight on the surface mesh. The currently implemeneted approaches only use the surface to determine which voxels to use for the analysis.

Also, my approach is dependent on a sphercial surface mesh, as for example FreeSurfer's `lh.sphere`. The vertex coordinates of this spherical mesh are needed to compute the neighborhood properties and therefore determine if a vertex is in the searchlight disk.

It's my first function implementation to PyMVPA. My approach was to add the `disk_searchlight`  everywhere where I also found a `sphere_searchlight`. I tried to add some tests that hopefully should verify that the new [`Disk`](https://github.com/PyMVPA/PyMVPA/compare/master...miykael:disk_searchlight?expand=1#diff-104766b285ac58e13da2dd02f2064776) class works correctly.

### DIFF
--- a/mvpa2/measures/gnbsearchlight.py
+++ b/mvpa2/measures/gnbsearchlight.py
@@ -14,7 +14,7 @@ __docformat__ = 'restructuredtext'
 import numpy as np
 
 from mvpa2.base.dochelpers import borrowkwargs, _repr_attrs
-from mvpa2.misc.neighborhood import IndexQueryEngine, Sphere
+from mvpa2.misc.neighborhood import IndexQueryEngine, Sphere, Disk
 
 from mvpa2.measures.adhocsearchlightbase import \
      SimpleStatBaseSearchlight, _STATS
@@ -23,7 +23,7 @@ if __debug__:
     from mvpa2.base import debug
     import time as time
 
-__all__ = [ "GNBSearchlight", 'sphere_gnbsearchlight' ]
+__all__ = [ "GNBSearchlight", 'sphere_gnbsearchlight', 'disk_gnbsearchlight' ]
 
 class GNBSearchlight(SimpleStatBaseSearchlight):
     """Efficient implementation of Gaussian Naive Bayes `Searchlight`.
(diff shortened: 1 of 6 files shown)


miykael:
To create a dataset for this approach, use `from mvpa2.datasets.gifti import gifti_dataset` and add the coordinates of the sphere vertices to the dataset, e.g.:
```python
from nibabel.freesurfer import read_geometry
coords, faces = read_geometry('lh.sphere')
ds.set_attr('fa.vertex_indices', coords)
```

The usage of `disk_searchlight` is afterwards almost the same as for `sphere_searchlight`:
```python
sl = disk_searchlight(cv,
                      radius=disk_radius,
                      center_ids=center_ids,
                      coords=ds.fa.vertex_indices,
                      space='vertex_indices')
```

Now, if you run `disk_searchlight` on every nth element, e.g. every 100th, you would get something as follows:
center_id_unequal

The distribution is unequal, because the vertices are not spaced equally. Therefore, I would recommend to use an additional function that spaces the searchlight disk center equally over the sphere. An example of such a function is a follows:
```python
def get_equally_spaced_points(coords, num_pts=100):
  """Return indices of vertices that cover the spherical mesh equally"""

  indices = np.arange(0, num_pts, dtype=float) + 0.5

  phi = np.arccos(1 - 2 * indices / num_pts)
  theta = np.pi * (1 + 5**0.5) * indices

  # Compute location of equally spaced points on a sphere
  x = np.cos(theta) * np.sin(phi)
  y = np.sin(theta) * np.sin(phi)
  z = np.cos(phi)

  # Compute sphere radius
  sphere_radius = np.mean(np.linalg.norm(coords, axis=1))

  # Adjust equally spaced points to requested sphere radius
  equal_points = np.vstack((x, y, z)).T * sphere_radius

  # Find vertices closest to `equal_points`
  equal_vertices = [np.argmin(np.linalg.norm(coords - e, axis=1))
                    for e in equal_points]
  return equal_vertices
```

This leads to an equal distribution of the search disks:
center_id_equal

mvdoc:
Hi @miykael,

maybe I'm misunderstanding something, but the functionality to run discal searchlights on the surface (so using data projected onto the surface) is already implemented in PyMVPA. For example, you can run a disc searchlight on the surface like this:

```python
# load surface
pial = surf.read(pial_fn)
# create query engine on the surface using dijkstra (~geodesic) distance
qe = SurfaceQueryEngine(pial, radius=10.0)

# set up searchlight
cvte = CrossValidation(
        LinearCSVMC(),
        partitioner,
        errorfx=mean_match_accuracy)
sl = Searchlight(cvte, queryengine=qe, nproc=nproc)

# run it
slmap = sl(ds)
```

nno:
Hi @miykael and @mvdoc,

@miykael thanks for the PR and our work, but unfortunately it seems we already have (most of) its functionality.
As @mvdoc wrote, PyMVPA supports both the surface-to-surface and surface-to-voxel approaches. For surface-to-surface the input dataset with data associated with each node has to be on the surface to, and should in the typical use case hava a field .fa.node_indices. For the surface-to-voxel approach (each  center node is associated with a set of neighbouring voxels), the input dataset has to be in volume space.

mvdoc:
@miykael maybe I'm not clear on what you're aiming at, but the current implemented methods offer the option to run

- volumetric searchlight
- surface to volume searchlight (using `SurfaceVerticesQueryEngine` for the method in Oosterhof et al., 2011)
- surface to surface searchlight (using `SurfaceQueryEngine`)

In the last case, your data need to be already projected to the surface.

As for the `node_indices` error, I usually just impute them `ds.fa['node_indices'] = np.arange(ds.nfeatures)`, if and only if the data is not masked. Otherwise you need to make sure of the correct correspondence between features and node indices.

miykael:
@nno - Thanks for the quick reply, too.

@mvdoc - Thanks for the detailed explanation. I'm also not clear on what I'm aiming at. I think I just somehow try to find a reason for my PR to be merged, nonetheless :-) But you're right the implemented methods cover already everything.

@nno and @mvdoc - Three years ago, I worked for a few weeks on this issue because I thought it wasn't implemented yet and the last global brainhack gave me now a motivational boost to finish it. But it seems I never really checked well enough :-) So, no worries. As it's already implemented, we can therefore close this PR.

yarikoptic:
it looks so lovely that it is painful to merge... if you think that anything could/should be repurposed somehow -- feel free to reopen. thanks @miykael

PR #585, opened 2018-05-11
Minor usecase which possibly should be absorbed within PDistTargetSimilarity or alike

yarikoptic:


mvpa2/tests/test_usecases.py
```diff
@@ -621,3 +621,67 @@ def test_permute_superord():
...
+
+    out = Dataset(np.empty((4, slres.nfeatures)))
+    out.sa["metric"] = ["corr", "p", "fisher(corr)", "z(p) rdist(15)"]
+    for ivoxel in range(self_vs_others.nfeatures):
+        corr, p = ss.spearmanr(
+            target_self_vs_others,
+            self_vs_others.samples[:, ivoxel])
+        out.samples[:, ivoxel] = [
+            corr, p, np.arctanh(corr), ssd.rdist(15).isf(p)]
+    print out.samples
```
nno: this can be removed?

mvpa2/tests/test_usecases.py
```diff
@@ -621,3 +621,67 @@ def test_permute_superord():
...
+
+    assert np.all(-1 <= out.samples[0])
+    assert np.all(out.samples[0] <= 1)
+    assert np.all(0 <= out.samples[1])
+    assert np.all(out.samples[1] <= 1)
+    # map2nifti(ds, out)
```
nno: also this can be removed I think

nno:
Looks good, just found two lines that seem leftovers during testing the testing that I think should be removed.

PR #588, opened 2018-06-26
Two fixes to make docs buildable again

yarikoptic:


yarikoptic:
sweet, so in 1.3.1 there were no yet `facecolor` so now we need to do the full dance of comparing matplotlib version... bleh

PR #591, opened 2018-11-15
[Work in progress] GNBWeights

AdinaWagner:
@yarikoptic: here is the mess I created so far! ;-) 

- class GNB has get_sensitivity_analyzer function that conditional on parameter common_variance=True creates and instance of GNBWeights
- there is a GNBWeights class now. It will take the difference of feature means in two classes and return them as a Dataset.
- there are some simply unit tests that assert whether classification works, an exception is raised as expected if sensitivity analyzer is called with a non-linear instance of gnb, the ``__tags__ `` thingy gets updated accordingly

I'm stuck particularly with: how do weights and sensitivity relate? a simple pointer to a paper/textbook/wikipedia page I wasn't able to find yet would suffice - I feel like this is a very basic question.
Some things about my GNBWeights class and get_sensitivity_analyzer() method don't make sense yet. While 'smlr.get_sensitivity_analyzer()' directly seems to run the _call method of SMLRWeights class and returns a dataset, gnb.get_sensitivity_analyzer() only returns a ``<GNBWeights>`` something. I don't know where the implementations of smlr and gnb differ to create this different behavior.

### DIFF
--- a/mvpa2/algorithms/connectivity_hyperalignment.py
+++ b/mvpa2/algorithms/connectivity_hyperalignment.py
@@ -145,7 +145,7 @@ class ConnectivityHyperalignment(SearchlightHyperalignment):
         seed_data = seed_data(dataset)
         zscore(seed_data, chunks_attr=None)
         return seed_data
-
+    
     def _get_sl_connectomes(self, seed_means, qe_all, datasets, inode, connectivity_mapper):
         # For each SL, computing connectivity of features to seed means
         sl_connectomes = []
@@ -158,7 +158,7 @@ class ConnectivityHyperalignment(SearchlightHyperalignment):
                 sl_ids = sl_ids.samples[0, :].tolist()
             sl_connectomes.append(connectivity_mapper.forward(sd[:, sl_ids]))
         return sl_connectomes
-
+    
     def _get_hypesvs(self, sl_connectomes, local_common_model=None):
         '''
         Hyperalign connectomes and return mapppers
@@ -203,7 +203,7 @@ class ConnectivityHyperalignment(SearchlightHyperalignment):
         params = self.params
         # If no precomputed connectomes are supplied, compute them.
         if params.connectomes is not None and os.path.exists(params.connectomes):
-            _chpaldebug("Loading pre-computed connectomes from ", params.connectomes)
+            _chpaldebug("Loading pre-computed connectomes from %s" % params.connectomes)
             connectomes = h5load(params.connectomes)
             return connectomes
         connectivity_mapper = FxyMapper(params.conn_metric)
--- a/mvpa2/base/__init__.py
+++ b/mvpa2/base/__init__.py
@@ -337,6 +337,7 @@ if __debug__:
     debug.register('HPAL',   "Hyperalignment")
     debug.register('HPAL_',  "Hyperalignment (verbose)")
     debug.register('SHPAL',  "Searchlight Hyperalignment")
+    debug.register('CHPAL',  "Connectivity Hyperalignment")
     debug.register('GCTHR', "Group cluster threshold")
     debug.register('ATL', "Atlases")
     debug.register('ATL_', "Atlases (verbose)")


yarikoptic:
> I'm stuck particularly with: how do weights and sensitivity relate? a simple pointer to a paper/textbook/wikipedia page I wasn't able to find yet would suffice - I feel like this is a very basic question.

Also from your email I see that terminology is confusing here.  First of all, the "sensitivity" we are talking here has nothing to do with the sensitivity of the classifier as a whole which is a synonym to the [True Positive Rate](https://en.wikipedia.org/wiki/Sensitivity_and_specificity).  The "sensitivity" of our interest is the question of how (much) classifier is sensitive to values in each feature, or in other words - what features are more important (higher sensitivity for them) and contribute more to decision than the others.  One of the most obvious and mentally understandable ways to assess such sensitivity is to start adding noise to each feature and see how it affects classification results. For that we have [NoisePerturbationSensitivity](www.pymvpa.org/generated/mvpa2.measures.noiseperturbation.NoisePerturbationSensitivity.html), but it is not a pragmatic approach.  The simplest one is to just look at the slope (the weights) of the decision boundary of a linear classifier (assuming that data first was normalized so that all features have similar variance across data) -- features with high weights are the ones which classifier relies most on while making its decision.  Changes in those features affect it most.  And features with small weights do not affect that much. Some times removing those not important features leads to greatly improved generalization, and that is a principle behind [Recursive Feature Elimination](http://www.pymvpa.org/generated/mvpa2.featsel.rfe.RFE.html).  See e.g. [Sensitivity Measure](www.pymvpa.org/examples/sensanas.html) demonstrating a few of such estimates on fMRI data.  In case of SVM, IIRC weights directly correspond to the derivative of the risk minimization objective function.
From another perspective -- it pretty much closely relates to any linear modeling approach, where upon fitting the model you check which of your variables have high contribution to explain the model ;-)

I hope this helps

> Some things about my GNBWeights class and get_sensitivity_analyzer() method don't make sense yet. While 'smlr.get_sensitivity_analyzer()' directly seems to run the _call method of SMLRWeights class and returns a dataset, gnb.get_sensitivity_analyzer() only returns a `<GNBWeights>` something. I don't know where the implementations of smlr and gnb differ to create this different behavior.

`GNBWeights` such as `SMLRWeights` is a class, so `get_sensitivity_analyzer()` must return an instance of it.  I do not see a direct call to `_call`: https://github.com/PyMVPA/PyMVPA/blob/master/mvpa2/clfs/smlr.py#L573

AdinaWagner:
just quick to clarify: 
The href of NoisePertubationSensitivity leads nowhere for me, is it meant to point here:
https://github.com/PyMVPA/PyMVPA/blob/master/mvpa2/measures/noiseperturbation.py ?

yarikoptic:
Try now ;-) was missing http:// prefix

AdinaWagner:
I hope you are battling your cold successfully. Let me know if you need a heroic German to bring a bottle of warm beer to Norwich ;-)

As an important first: I will stop thinking about this until at least tomorrow - instead I'm going to hopefully drama-free move and have a beer to celebrate it afterwards. Disregard this, I just leave these thoughts down here as a reminder of where I was. If you're here tomorrow and I haven't encountered a revelation via dream I might ask you stuff, if you have time. Else I will be in Washington DC Saturday morning and will ask a mathematician about my weird math questions. 

I'm still managing to confuse myself. I'm having small moments of enlightenment, but not enough to come up with sensible procedural steps to undertake. Its not so much of trouble implementing something (because I'm not even there yet), but trouble with grasping/using the underlying mathematical properties of the GNB algorithm. I will try to outline my understanding of everything in very small steps. Ideally, you can put a mental check mark after any correct statement and get an idea of where and how I am wrong. I try to use sams analysis as a handrail.

- For the analysis we have 7 classes/labels (FFA, PPA, EV, LOC, EBA, OFA, rest of brain). 1, ..., k

- We have ~125.000 voxels per brain, i.e. 125.000 samples x 1, ..., i.

- There are 3599 time points. The time points are the features 1, ... j.

- The objective is to label an individual voxel (sample x_new) based on its characteristics on all features with one class label (i.e. classify which brain region it belongs to)

- A Gaussian Naive Bayes classifier will for every sample x_i compute the probability for all of the classes given the sample, p(k | x_i). Therefore, for each sample, it computes 7 probabilities. The sample will get the label of class k with the highest probability p(k|x_i) of all 7 probabilities. 
(i.e. codecogseqn-2 )

- For the probability computation, the GNB will during testing model the distribution of values for each feature given each class as a Gaussian distribution. I.e. it will compute the parameters of the distribution of each feature for each of the 7 classes. As we want to set common_variance=True, I assume it will only model the mean of each feature, which means taking the average activation of all voxels (samples) on one timepoint (for one feature) that belong to a specific class k.

- As for classification during testing, the decision of "which class does an unseen voxel belong to" boils down to the sum of squared distances of the new data point on every feature given each class to the mean of each 'feature given class' distribution mean in this thing:  
codecogseqn-1
and then multiplying the resulting probabilities per class. 

- However, the weighting of each feature is not determined during testing, but during training

- The idea behind sensitivity analysis is to find out which features (time points) are especially useful in classifying a datapoint. Those timepoints that can't contribute much to classification should be disregarded to improve the performance of the classifier.

- The weight for a feature in a GNB sounds to me so far to just be the conditional probability P(x|k) that is put into the Bayes rule as a class model
codecogseqn-3

- this is however class- and feature-specific - I have one weight per feature per class, I don't have a feature specific weight

- all examples of classification with gnb I have found so far use examples with binary decisions. it has to be possible to have more than two classes labeled. The generalization of the formulas to the case of three or more classes is termed "trivial" but its not very trivial without having studied maths...


... i will call it a day now - stupid knot in my brain.

AdinaWagner:
@yarikoptic - tests pass! I will need to add a some more elaborate tests, I suppose. Could you take a quick look once you get a chance and see whether everything looks kosher?

mvpa2/clfs/gnb.py
```diff
@@ -289,36 +299,58 @@ def _predict(self, data):
...
+        # get means of all attributes given class label
+        means = clf.means
+        # number of features
+        nfeat = clf.means.shape[1]
+
+        # make sure labels are numeric
+        am = AttributeMap()
+        ulab = am.to_numeric(clf.ulabels)
+        # all pairwise combinations of labels
+        lcomb = list(itertools.combinations(ulab, 2))
```
yarikoptic: I think I was wrong pointing you to the `AttributeMap` -- no need here. Just work with the indices as they correspond to the ulabels: `itertools.combinations(range(len(clf.ulabels)))` and then when you want a real label, just `clf.ulabels[idx]`

mvpa2/clfs/gnb.py
```diff
@@ -289,36 +299,58 @@ def _predict(self, data):
...
+
+        weights = np.zeros([len(lcomb), nfeat])
+        # do not compute sensitivity for features with variance 0 as this would implicate
+        # a division by zero
+        zero_vars = clf.variances==0
+        for pair in lcomb:
+            for i in range(0, nfeat):
```
yarikoptic: just rely on numpy's ability to work with vectors. So here you really want just
`python
weights[row_idx] = (means[pair[0]] - means[pair[1])/stddev
`

mvpa2/clfs/gnb.py
```diff
@@ -289,36 +299,58 @@ def _predict(self, data):
...
+        am = AttributeMap()
+        ulab = am.to_numeric(clf.ulabels)
+        # all pairwise combinations of labels
+        lcomb = list(itertools.combinations(ulab, 2))
+
+        weights = np.zeros([len(lcomb), nfeat])
+        # do not compute sensitivity for features with variance 0 as this would implicate
+        # a division by zero
+        zero_vars = clf.variances==0
+        for pair in lcomb:
```
yarikoptic: just include your index in your "looping" setup:
``python
for pair_idx, pair in enumerate(lcomb):
```

File: mvpa2/clfs/gnb.py [hunk 4]
```diff
@@ -289,36 +299,58 @@ def _predict(self, data):
...
+                var_idx = np.where(ulab == int(pair[0]))[0][0]
+                var = clf.variances[var_idx, i]
+                if zero_vars[var_idx, i]:
+                    continue
+                row_idx = lcomb.index(pair)
+                weights[row_idx, i] = (means[int(pair[0]), i] - means[int(pair[1]),i])/var
+
+        # put everything into a Dataset
+        ds = Dataset(weights,
+                     sa={clf.get_space(): list(itertools.combinations(clf.ulabels, 2))})
```
yarikoptic:
```
[(clf.ulabels[p1], clfs.ulabels[p2]) for p1, p2 in lcomb]
```
recommend renaming lcomb into `pairs` to make it easier to read

mvpa2/tests/test_gnb.py
```diff
@@ -55,6 +56,38 @@ def test_gnb(self):
...
+                            gnb_.get_sensitivity_analyzer()
+
+
+def test_gnb_sensitivities():
+    gnb = GNB(common_variance=True)
+    ds = normal_feature_dataset(perlabel=4,
+                                nlabels=3,
+                                nfeatures=5,
+                                nchunks=4,
+                                snr=10
```
yarikoptic: Please add this so there is signal added to those 3 features
```python
nonbogus_features=[0, 1, 2]
```
this would add signal for first label into feature index 0, 2nd label into feature index 1 and so on, so later on you could actually see the effect on sensitivies.

yarikoptic:
Someone got into force passing things and started unintentionally closing PRs?

PR #593, opened 2018-12-03
Connectivity Hyperalignment log issue

anibalsolon:
Some minor error fixes for CHA when running with `MVPA_DEBUG=CHPAL.*`.


yarikoptic:
Ha ha -- thanks!

PR #595, opened 2018-12-14
[FIX]: First attempt to get a classifiers plot function also build confusion matrices from a subset of labels

AdinaWagner:
This PR:
- allows plotting of confusion matrices with a subset of labels of the original classification, if they are specified as strings within the ``labels`` parameter
- disables plot() from blowing up if the length of provided labels does not match the dimensions of the matrix
- add a small smoketest to see whether plotting of full or subsets of the labels blows up

### DIFF
--- a/mvpa2/clfs/transerror.py
+++ b/mvpa2/clfs/transerror.py
@@ -969,6 +969,10 @@ class ConfusionMatrix(SummaryStatistics):
                     v = None
                     if l is not None: v = labels_map_full[l]
                     labels_plot += [v]
+            elif (len(labels) != labels_order_filtered_set) and \
+                    all(isinstance(x, str) for x in labels_order_filtered_set):
+                # get those items in provided label parameter that correspond to the existing labels
+                labels_plot = [label for label in labels_order_filtered_set if label in labels]
             else:
                 raise ValueError, \
                       "Provided labels %s do not match set of known " \
@@ -984,10 +988,13 @@ class ConfusionMatrix(SummaryStatistics):
         NlabelsNN = len(non_empty)
         Nlabels = len(labels_plot)
 
-        if matrix.shape != (NlabelsNN, NlabelsNN):
-            raise ValueError, \
-                  "Number of labels %d doesn't correspond the size" + \
-                  " of a confusion matrix %s" % (NlabelsNN, matrix.shape)
+        # I believe this Error is superfluous -- or rather: wouldn't let
+        # subsets of matrices be printed.
+
+        #if matrix.shape != (NlabelsNN, NlabelsNN):
+        #    raise ValueError, \
+        #          "Number of labels %d doesn't correspond the size" + \
+        #          " of a confusion matrix %s" % (NlabelsNN, matrix.shape)
 
         confusionmatrix = np.zeros((Nlabels, Nlabels))
         mask = confusionmatrix.copy()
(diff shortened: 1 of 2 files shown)


AdinaWagner:
(@yarikoptic)

mvpa2/clfs/transerror.py
```diff
@@ -969,6 +969,10 @@ def plot(self, labels=None, numbers=False, origin='upper',
                     v = None
                     if l is not None: v = labels_map_full[l]
                     labels_plot += [v]
+            elif (len(labels) != labels_order_filtered_set) and \
+                    all(isinstance(x, str) for x in labels_order_filtered_set):
```
yarikoptic: type shouldn't matter really - I might be having numeric labels
also, why not just to use sets operation to check exactly that `labels_order_filtered_set` includes all we would like to plot like
```suggestion
            elif not (set(labels) - labels_order_filtered_set):
```
? because `len` is too weak of a check -- I could have completely different labels from what is provided to labels_order and condition would still pass, but then you would end up with no `labels_plot`

mvpa2/clfs/transerror.py
```diff
@@ -984,10 +988,13 @@ def plot(self, labels=None, numbers=False, origin='upper',
         NlabelsNN = len(non_empty)
         Nlabels = len(labels_plot)
 
-        if matrix.shape != (NlabelsNN, NlabelsNN):
-            raise ValueError, \
-                  "Number of labels %d doesn't correspond the size" + \
-                  " of a confusion matrix %s" % (NlabelsNN, matrix.shape)
+        # I believe this Error is superfluous -- or rather: wouldn't let
+        # subsets of matrices be printed.
```
yarikoptic: do we want subsets of matrices printed?  that is related to the comment above.  It could be a feature indeed, but I think if desired, it should be explicit, like adding an additional parameter `allow_partial=False`, so by default we would blow (so not to miss errors in specification) and only if explicitly instructed -- then to be allowed.

mvpa2/tests/test_transerror.py
```diff
@@ -673,6 +673,26 @@ def test_confusionmatrix_nulldist(self):
                                       cvnp[(np.array([0,1]), np.array([1,0]))])
 
 
+def test_plotting_subset():
```
yarikoptic: so if we do allow (optionally) the subset to be visualized -- then yeah, worth having this test

our use case though is different -- it is when we provide labels (to order) which has more labels than it is known to the ca.stats.labels (like we include "brain" in providing the order, but then we had no "brain" label within our dataset we worked on)

mvpa2/tests/test_transerror.py
```diff
@@ -673,6 +673,26 @@ def test_confusionmatrix_nulldist(self):
...
+    ds = normal_feature_dataset(nlabels=4)
+    clf = GNB()
+    cv = CrossValidation(
+        clf, NFoldPartitioner(),
+        errorfx=None,
+        enable_ca=['stats'])
+    res = cv(ds)
+    # get all labels
+    labels = cv.ca.stats.labels
+    cv.ca.stats.plot(labels=labels, numbers=True)
```
yarikoptic: for unit testing `ConfusionMatrix` there is no need to do full blown classification.  Just create an instance of ConfusionMatrix (see other tests) and work with that.

PR #525, opened 2017-05-31
RF: Python 2/3 compatibility

chrispycheng:
making code py2/py3 compatible per yarikoptic's instructions. please lmk if I did anything procedurally incorrectly.

TODOs (@yarikoptic's edits):
- [x] establish a baseline with some very basic unittests passing on both py2 and py3
- [ ] keep going through components and mark them when corresponding submodules and tests are
      converted.
      While at it, I will try to move tests around (possibly screwing up some manual collection of tests) so we could adhere to the same "locality" setup as we do in other projects
     - [x] `base/` done in the sense that you can run tests on it, but not completely:
          - [ ] There is some discrepancy in summary output stats for some datasets between py2 and py3 - so something to be fixed up and solidified with a test. 
          - [ ] `test_hdf.py` for now resides under main `tests/` since pulls in too many other submodules
     - [ ] ...
- [x] reincarnate coverage reporting (was disabled for now)
  - [ ] there is still some pieces commented out TODO to reenable coverage
- [ ] compare performance hit (against `master`) for py2.  In many places we just use `.items()`, `.keys()` etc -- judging from timings of the run on travis -- there is sever hit ATM, but may be it is solely because we have more external dependencies installed for py2 than py3

PR #601, opened 2019-03-20
Convert bytes output of subprocess to string for Python 3

### LINKED ISSUE #600: Subprocess gives TypeError when running pymvpa2-tutorial in Python3
```bash
> ./pymvpa2-tutorial --tutorial-data-path=/Users/justinshenk/Downloads/tutorial_data       
Use existing tutorial data path at '/Users/justinshenk/pymvpa2_tutorial/data'.
  File "./pymvpa2-tutorial", line 172, in <module>
    main()
  File "./pymvpa2-tutorial", line 155, in main
    ipython_ver, ipython = get_latest_ipython()
  File "./pymvpa2-tutorial", line 75, in get_latest_ipython
    l = p.stdout.read().split('\n')
TypeError: a bytes-like object is required, not 'str'
```

justinshenk:
Resolves #600.

### DIFF
--- a/bin/pymvpa2-tutorial
+++ b/bin/pymvpa2-tutorial
@@ -69,7 +69,7 @@ def get_latest_ipython(ipythons=['ipython01x', 'ipython']):
             p = subprocess.Popen([ipython, '-V'],
                                  stdout=subprocess.PIPE,
                                  stderr=subprocess.PIPE)
-            l = p.stdout.read().split('\n')
+            l = str(p.stdout.read()).split('\n') # convert to string for Python 3
             l = [x for x in l if len(x)]    # strip empty lines
             if len(l):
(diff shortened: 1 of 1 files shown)


justinshenk:
Codecov references seemingly unrelated code https://codecov.io/gh/PyMVPA/PyMVPA/pull/601/src/mvpa2/tests/test_searchlight.py.

yarikoptic:
Thanks - I will merge it since it shouldn't make things worse ;)
but note that python3 is not yet officially supported. Even though pymvpa2 would install and seems to function correctly, it might not actually be doing the right thing - tests would fail in you run them. An overall overhaul of the codebase to use python-future to support both 2 and 3 is slowly ongoing in #525

PR #602, opened 2019-03-20
BF: set affine to the one we know in map2nifti

### LINKED ISSUE #599: map2nifti returns an object with no affine transformation
Hi all, I noticed today that map2nifti appears to not set the affine property of the Nifti1Image object. It appears to be by design in that [line 181 of mri.py](https://github.com/PyMVPA/PyMVPA/blob/327d282142a9e80080e6d5bd9ced6652a96bd954/mvpa2/datasets/mri.py#L181) explicitly sets the affine to None which in turn results in nibabels's spatialimages.SpatialImage setting the obj.affine to none as well (as per the docstring).  

Is there any danger in map2nifti passing the ds.a.affine attribute on to nibabel so that it's set in the niftimage object? I'm not understanding the decision to explicitly set it to None at line 181. 

If you're wondering why I'm asking, I came across this while using nilearn's plotting to plot a PyMVPA dataset. Nilearn expects the Nifti1Image obj to have an affine property but map2nifti omits it. The workaround is to map2nifti.to_filename and then load the nifti file with nilearn, but I'd like to skip the intermediate step! 

Cheers.

yarikoptic:
Original comment hinted that it was supposed to be set by nibabel if provided
affine is None.  I have tested in precise docker image with nibabel 1.1.0 (2011ish)
and it (already) was not the case.  So better to set it explicitly and test for
that in the tests.  Thanks @ddwagner for the report and analysis!

Closes #599

### DIFF
--- a/mvpa2/datasets/mri.py
+++ b/mvpa2/datasets/mri.py
@@ -176,9 +176,8 @@ def map2nifti(dataset, data=None, imghdr=None, imgtype=None):
     if issubclass(imgtype, nibabel.spatialimages.SpatialImage) \
             and (imghdr is None or hasattr(imghdr, 'get_data_dtype')):
         # we can handle the desired image type and hdr with nibabel
-        # use of `None` for the affine should cause to pull it from
-        # the header
-        return imgtype(_get_xyzt_shaped(dsarray), None, imghdr)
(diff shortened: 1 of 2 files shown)


yarikoptic:
The PR is small, so if side-effects are detected - we will just revert

PR #606, opened 2019-03-28
Set division in HalfPartitioner to floor division for Python 3 compat

ddwagner:
Running through the tutorial with students noticed that the HalfPartitioner throws an error in Python3 but not 2. The culprit is the change to division in Python 3 that now returns float instead of int. Integer divsion (//) works in Python 2 so shouldn't break compatibility.

### DIFF
--- a/mvpa2/generators/partition.py
+++ b/mvpa2/generators/partition.py
@@ -319,8 +319,8 @@ class HalfPartitioner(Partitioner):
         list of tuples (None, list of int)
           2 items: first half of samples into 1st split
         """
-        return [(None, uniqueattrs[:len(uniqueattrs)/2]),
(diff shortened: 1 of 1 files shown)


yarikoptic:
Oh. Better not even try to use it with python 3 ATM. Eventually https://github.com/PyMVPA/PyMVPA/pull/525 will get finished

ddwagner:
Ahh dang. I thought Python 3 was basically "at your own risk" but mostly working. Hmm. I'll either drop to 2.7 or use the NGroupPartitioner until we run into the next Python 3 issue!

yarikoptic:
well -- it is "mostly" working but not necessarily correctly ;)

PR #607, opened 2019-05-04
A variety of fixes to make target pairs consistently object array of tuples with (neg, pos) targets

yarikoptic:
- GNB had sensitivity negated
- make GNB `has_sensitivity` added in constructor so we discover/test it from the warehouse
- test that MulticlassClassifier's sensitivity would match the one produced by the classifier directly
  - make order of targets for BinaryClassifier also be (neg, pos)
- make libsvm's SVM produce tuples of pairs even when no attrmap was assigned (SG one was doing that already)
- more ...

### DIFF
--- a/mvpa2/clfs/gnb.py
+++ b/mvpa2/clfs/gnb.py
@@ -121,6 +121,11 @@ class GNB(Classifier):
         # Define internal state of classifier
         self._norm_weight = None
 
+        # Add 'has_sensitivity' tag if classifier is linear
+        if self.params.common_variance:
+            self.__tags__ = self.__tags__ + ['has_sensitivity']
+
+
     def _get_priors(self, nlabels, nsamples, nsamples_per_class):
         """Return prior probabilities given data
         """
@@ -203,11 +208,7 @@ class GNB(Classifier):
         else:
             self._norm_weight = 1.0/np.sqrt(2*np.pi*variances)
 
-        # Add 'has_sensitivity' tag if classifier is linear
-        if params.common_variance \
-            and 'has_sensitivity' not in self.__tags__:
-            self.__tags__ += ['has_sensitivity']
-
+        assert params.common_variance == ('has_sensitivity' in self.__tags__)
 
         if __debug__ and 'GNB' in debug.active:
             debug('GNB', "training finished on data.shape=%s " % (X.shape, )
@@ -222,9 +223,6 @@ class GNB(Classifier):
         self.ulabels = None
         self.priors = None
         super(GNB, self)._untrain()
-        # Remove 'has_sensitivity' tag
-        if 'has_sensitivity' in self.__tags__:
-            self.__tags__.remove('has_sensitivity')
 
 
     @accepts_dataset_as_samples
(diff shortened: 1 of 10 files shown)


yarikoptic:
ok, I think there is no objections

mvpa2/clfs/gnb.py
```diff
@@ -332,16 +331,26 @@ def _call(self, dataset):
...
+        nonzero_vars0 = nonzero_vars[0, :]
         for idx, pair in enumerate(pairs):
-            weights[idx, nonzero_vars[0, :]] = (means[pair[0], nonzero_vars[0, :]] -
-                                                  means[pair[1], nonzero_vars[0, :]]) / \
-                                                 clf.variances[pair[0], nonzero_vars[0, :]]
+            # two-class sensitivity for (L0, L1) assumes that L1 is the
+            # "positive one"
+            weights[idx, nonzero_vars0] = (means[pair[1], nonzero_vars0] -
+                                           means[pair[0], nonzero_vars0]) / \
+                                          clf.variances[pair[0], nonzero_vars0]
```
adswa: @yarikoptic I'm double checking sign of resulting beta estimates. Could you check whether I think correct here, just so that I'm not missing something obvious?

Situation: GLM on sensitivities (FFA-PPA) yields negative values for faces (as in plot seen before).

``pairs = list(itertools.combinations(range(len(clf.ulabels)), 2))`` will yield tuples [emitted in lexicographic sort order](https://docs.python.org/2/library/itertools.html), i.e. will have [('FFA', 'PPA'), ('EBA', 'FFA'), ...] instead of [('PPA', 'FFA'), ('FFA', 'EBA'), ...] and so forth.

```
weights[idx, nonzero_vars0] = (means[pair[1], nonzero_vars0] -
                               means[pair[0], nonzero_vars0]) / \
                               clf.variances[pair[0], nonzero_vars0]
```
will in the case of the ('FFA', 'PPA') pairs hence subtract FFA (index 0) estimates from PPA (index 1) estimates. If the FFA has a higher activation (mean) in a feature in question (e.g. faces) than the PPA, this would result in a negative feature weight for faces in the FFA-PPA decision boundary. 
Is that correct or am I missing something obvious?

yarikoptic: > pairs = list(itertools.combinations(range(len(clf.ulabels)), 2)) will yield tuples emitted in lexicographic sort order, i.e. will have [('FFA', 'PPA'), ('EBA', 'FFA'), ...] instead of [('PPA', 'FFA'), ('FFA', 'EBA'), ...] and so forth.

note the `range` in original invocation.  So here is no labels, just indexes to the labels.  `.ulabels` though should indeed be sorted but I would not assume it.  So if you have two labels

```
In [6]: list(itertools.combinations(range(2), 2))
Out[6]: [(0, 1)]
```
and if you have 3:
``
In [5]: list(itertools.combinations(range(3), 2))
Out[5]: [(0, 1), (0, 2), (1, 2)]
```
so you can see that the second index growing 'faster' so I think that neither of the two of your examples are correct ;) "lexicographic sort order" means here that those pairs are ordered in the order of original position(s) not that they are sorted alphabetically within entries:
```
In [7]: list(itertools.combinations('ABC', 2))
Out[7]: [('A', 'B'), ('A', 'C'), ('B', 'C')]

In [8]: list(itertools.combinations('CBA', 2))
Out[8]: [('C', 'B'), ('C', 'A'), ('B', 'A')]
```

adswa: ahh crap, thanks for helping :)

PR #610, opened 2019-05-17
BF: guard against under/overflows and mitigate in case of logprob=True (default)

### LINKED ISSUE #581: Wrong GNB probabilities
Reported on the mailing list by Richard Dinga

```
Thanks, this seems to be the solution according to documentation, but it doesn't work for my data. Here is an example using tutorial data:

from mvpa2.suite import *
datapath = '/usr/share/data/pymvpa2-tutorial/'
haxby = load_tutorial_data(datapath,
        roi='vt',
        add_fa={'vt_thr_glm': os.path.join(datapath, 'haxby2001',
                                           'sub001', 'masks',
                                                     'orig', 'vt.nii.gz')})
poly_detrend(haxby, polyord=1, chunks_attr='chunks')
haxby = haxby[np.array([l in ['rest', 'house', 'face']
                           for l in haxby.targets], dtype='bool')]
zscore(haxby, chunks_attr='chunks', param_est=('targets', ['rest']), 
       dtype='float32')
haxby = haxby[haxby.sa.targets != 'rest']
haxby = remove_invariant_features(haxby)

clf = GNB(enable_ca='estimates', logprob=True, normalize=True)
cv = CrossValidation(clf, NFoldPartitioner(attr='chunks'), postproc=None)
cv_results = cv(haxby)
print clf.ca.estimates

[[          inf           inf]
 [-234.34792494    0.        ]
 [          inf           inf]
 ..., 
 [          inf           inf]
 [          inf           inf]
 [          inf           inf]]

However it works with test data:

from mvpa2.testing.datasets import *
ds_test = datasets['uni2medium']
cv_results = cv(ds_test)
print np.round(np.exp(clf.ca.estimates), 3)

[[ 0.956  0.044]
 [ 1.     0.   ]
 [ 1.     0.   ]
 ..., 
 [ 0.     1.   ]
 [ 0.168  0.832]
 [ 0.001  0.999]]
```

yarikoptic:
we ~~mitigate~~ try to avoid it by offseting all the values by the maximal value (in log space),
computing marginals, and then offsetting back when back in the logspace.

Similar strategy should be possible for operating in original, non logspace.
But may be there is even a better way, so not bothering for now.

Attn @dinga92 , sorry it took so long ;)

Also the most recent change by @adswa  who introduced the test for GNB sensitivites with an evil scenario of no variance in one of the features, and `MVPA_DEBUG=ENFORCE_CA_ENABLED` triggering it lead to another fix/workaround for those cases.

Closes #581

### DIFF
--- a/mvpa2/clfs/gnb.py
+++ b/mvpa2/clfs/gnb.py
@@ -25,6 +25,7 @@ import numpy as np
 from numpy import ones, zeros, sum, abs, isfinite, dot
 from mvpa2.base import warning, externals
 from mvpa2.clfs.base import Classifier, accepts_dataset_as_samples
+from mvpa2.base.learner import DegenerateInputError
 from mvpa2.base.types import asobjarray
 from mvpa2.base.param import Parameter
 from mvpa2.base.state import ConditionalAttribute
@@ -206,6 +207,12 @@ class GNB(Classifier):
             variances[:] = cvar
         else:
             variances[non0labels] /= nsamples_per_class[non0labels]
+
+        if len(np.unique(means)) <= 1 and len(np.unique(variances)) <= 1:
+            raise DegenerateInputError(
+                "All means and variances are identical, cannot train GNB to "
+                "produce meaningful results"
+            )
 
         # Precompute and store weighting coefficient for Gaussian
         if params.logprob:
--- a/mvpa2/clfs/gnb.py
+++ b/mvpa2/clfs/gnb.py
@@ -101,6 +101,12 @@ class GNB(Classifier):
              disabled by default since does not impact classification output.
              """)
 
+    guard_overflows = Parameter(True, constraints='bool',
+            doc="""Computation of marginals could experience under and overflows
+            causing NaNs and Infs to emerge.  When enabled, GNB will verify
+            having finite numbers and mitigate the issue while computing
+            (ATM only in logprob=True mode). """)
+
     def __init__(self, **kwargs):
         """Initialize an GNB classifier.
         """
@@ -230,6 +236,7 @@ class GNB(Classifier):
         """Predict the output for the provided data.
         """
         params = self.params
+        guard_overflows = params.guard_overflows
         # argument of exponentiation
         scaled_distances = \
             -0.5 * (((data - self.means[:, np.newaxis, ...])**2) \
@@ -256,7 +263,8 @@ class GNB(Classifier):
 
             # Incorporate class probabilities:
             prob_cs_cp = lprob_cs + np.log(self.priors[:, np.newaxis])
-
+            if guard_overflows:
+                assert np.all(np.isfinite(lprob_cs))
         else:
             # Just a regular Normal distribution with per
             # feature/class mean and variances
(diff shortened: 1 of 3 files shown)


yarikoptic:
@dinga92 note that probabilities, as # of features grow, quickly quantize into 0, 1 values so most likely not really useful for anything really ;)

There might be a more scientific explanation but here is what I came up with for that matter:

Pretty much it boils down to diminishing probability density value for a possible point at the same distance from the mean as the dimensionality grows (you can do mentally comparison of 1d vs 2d or 3d scenario).  It just becomes probable to end up in many more points away from the                                                                                                                                                                 
decision boundary.
e.g. mentally imagine a bar (1d), ribbon (2d) or slab (3d) of the same thickness around point (1d), line (2d), plane (3d) where probabilities are equal (the same distance from means when assuming equal variance).  Now compare % of "space" occupied by outside of that bar/ribbon/slab.   you will note that in 2d case already it is much more space than in 1d. And in 3d MUCH more ;)                                                                                                                                                                                                                                   

So at the end, probability itself becomes really close to 0 or 1 in case of large number of dimensions and probably not useful much besides binary decision making.

In other classifiers, e.g. libsvm, "probability" is estimated based on some fit based on the distance to the decision boundary (so it is not a probability of belonging to the class, as in GNB), and should not be (guess) affected by the # of features, so could be used as a diagnostic.

dinga92:
@yarikoptic exploding of probabilities to 0-1 is a known issue of GNB :(, it happens because it's naive, so each voxel is considered as independent evidence for the class of the sample and of course if you have couple of voxels "independently" predicting one class, the probability soon blows.

dinga92:
However, this shouldn't be a problem with small searchlights. Is there a way to get GNBSearchlight to return probabilities? I think the new feature works only on gnb classifier, but not on SL right?

I tried to change the code by exponentiating lprob_cs_cp_sl in GNBSearchlight _sl_call_on_a_split but these seems not to sum to 1, but it can also be because I don't know what I am doing. Any ideas how to do that?

dinga92:
I am really sorry for spamming. I think I solved it for SL by adding normalization by marginals according to what is done in GNB classifier. It seems to work fine for me now (for sl size 1). 

I just added these lines

    prob_cs_cp_sl_real = np.exp(lprob_cs_cp_sl)
    prob_marginals = np.sum(prob_cs_cp_sl_real, axis=0)
    prob_predictions = prob_cs_cp_sl_real/prob_marginals
    prob_predictions = prob_predictions[1,:,:]

These are now real probabilities, right?

yarikoptic:
Sorry, I am away from this code atm.  But it is not clear to me why you need to change anything in the code.  With all the fixes here `.ca.estimates` should be the probabilities (or their log if `logprob=True`) if you have `normalize=True`

dinga92:
I couldn't think of a better way to do it. When I try to do it I just get `AttributeError: 'ConditionalAttributesCollection' object has no attribute 'estimates'` As far as I can see `sphere_gnbsearchlight` does not have the ability to output probabilities with ca or with another way.

yarikoptic:
because those are .ca of the classifier itself, not of a searchlight.

PR #614, opened 2019-08-11
BLD: for travis, set language to java and use trusty as distribution

nno:
This PR tries to address recent travis build errors, which started at

https://travis-ci.org/PyMVPA/PyMVPA/builds/545881884

with commit 

https://github.com/PyMVPA/PyMVPA/commit/2617c4d8dfe3fbe9e53f8e703ea996ab937b5c8f

although the root cause of the failing tests is probably a change in the default Ubunto configuration used on travis rather than code changes in PyMVPA.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@
 # travis-ci; it uses virtualenvs, they do not have numpy, scipy, matplotlib,
 # and it is impractical to build them
 language: java
+dist: trusty
 cache:
   - apt
 env:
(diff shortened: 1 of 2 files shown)


yarikoptic:
Thanks @nno for looking into it!!!

nno:
It seems a rounding issue is causing another test to fail (I may have written this test). I will try to look into this tomorrow.

nno:
@yarikoptic it seems travis is happy now. Anything stopping you from clicking on that shiny green button? :-)

yarikoptic:
Thanks!

PR #633, opened 2022-10-24
Transform errors into accuracies

sajjadtorabian:
Small issue, but probably important if consistency wants to be maintained with volumetric searchlight.py

### DIFF
--- a/doc/examples/searchlight_surf.py
+++ b/doc/examples/searchlight_surf.py
@@ -292,6 +292,12 @@ visualization. Finally print an informative message on how the
 generated data can be visualized using SUMA.
 """
 
+"""
+To maintain consistency with doc/examples/searchlight.py we transform the error maps into accuracies.
+"""
+sl_dset.samples *= -1
+sl_dset.samples += 1
+
 # save as NIML dataset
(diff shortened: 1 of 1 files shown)


sajjadtorabian:
Forgot to tag you, @yarikoptic

doc/examples/searchlight_surf.py
```diff
@@ -292,6 +292,12 @@
 generated data can be visualized using SUMA.
 """
 
+"""
+To maintain consistency with doc/examples/searchlight.py we transform the error maps into accuracies.
+"""
+sl_dset.samples *= -1
```
yarikoptic, line 298: why do you think they are errors? at https://github.com/PyMVPA/PyMVPA/pull/633/files#diff-97bb4eaabb66f937e134cb1c991de220b060f8b6e7f3990ed881639c8f3ee96fR193 I see that it is `np.mean(p == t)` so should be accuracy.

sajjadtorabian: Ah, that makes sense. The CrossValidation functions in searchlight.py and seachlight_surf.py differ in their errorfx argument (which is set to accuracy in surf as you mentioned) and I had missed that. Thanks for clarifying

